PR #2, opened 2020-03-19
Feat/met 2539 order xml records

acerdas84:
Main changes
src/main/java/eu/europeana/metis/sandbox/consumer/workflow/CreatedConsumer.java
src/main/java/eu/europeana/metis/sandbox/service/workflow/OrderingServiceImpl.java
src/main/java/eu/europeana/metis/sandbox/config/SandboxConfig.java

Other changes include:
Fix rabbit dlq 
Create Event record to transport records between steps
Set the application to create a war instead of a jar


src/main/java/eu/europeana/metis/sandbox/common/amqp/RecordMessageConverter.java
```diff
@@ -23,17 +24,19 @@
...
   @Override
   public Message toMessage(Object object, MessageProperties messageProperties)
       throws MessageConversionException {
-    if (!(object instanceof Record)) {
+    if (!(object instanceof Event)) {
       throw new MessageConversionException("Provided object is not of type Record");
     }
 
-    Record record = (Record) object;
+    Event<Record> recordEvent = (Event<Record>) object;
```
jochen-vermeulen: Perhaps you can change this to 
`Event<?> recordEvent = (Event<?>) object;`
So that we don't get an 'unchecked cast' warning.

src/main/java/eu/europeana/metis/sandbox/common/amqp/RecordMessageConverter.java
```diff
@@ -23,17 +24,19 @@
   private static final String COUNTRY = "country";
   private static final String STATUS = "status";
   private static final String STEP = "step";
+  private static final String EXCEPTION = "exception";
 
   private static final Charset DEFAULT_CHARSET = StandardCharsets.UTF_8;
 
   @Override
   public Message toMessage(Object object, MessageProperties messageProperties)
       throws MessageConversionException {
```
jochen-vermeulen: This exception is a RuntimeException and does not need to be explicitly mentioned here?

src/main/java/eu/europeana/metis/sandbox/domain/Event.java
```diff
@@ -0,0 +1,59 @@
+package eu.europeana.metis.sandbox.domain;
+
+import static java.util.Objects.requireNonNull;
+
+import eu.europeana.metis.sandbox.common.Status;
+import eu.europeana.metis.sandbox.common.Step;
+import org.apache.commons.lang3.exception.ExceptionUtils;
+
+public class Event<T extends Record> {
```
jochen-vermeulen: Is generic parameter T ever anything other than equal to Record? Do we need to apply generics here?

src/main/java/eu/europeana/metis/sandbox/config/SandboxConfig.java
```diff
@@ -1,12 +1,19 @@
 package eu.europeana.metis.sandbox.config;
 
+import eu.europeana.metis.transformation.service.TransformationException;
+import eu.europeana.metis.transformation.service.XsltTransformer;
 import eu.europeana.metis.utils.ZipFileReader;
+import java.io.File;
+import java.io.IOException;
+import java.net.URL;
```
jochen-vermeulen: Unused imports.

src/main/java/eu/europeana/metis/sandbox/config/SandboxConfig.java
```diff
@@ -24,12 +31,30 @@
   @Value("${sandbox.dataset.creation.threads.thread-prefix}")
   private String threadPrefix;
 
+  @Autowired
+  private ResourceLoader resourceLoader;
```
jochen-vermeulen: It is recommended to not use field injection, but rather @autowired constructors. This is so that you can mock them better.

src/main/java/eu/europeana/metis/sandbox/config/amqp/CreatedConfiguration.java
```diff
@@ -21,6 +21,9 @@
   @Value("${sandbox.rabbitmq.queues.record.created.dlq}")
   private String dlq;
 
+  @Value("${sandbox.rabbitmq.exchange.dlq}")
+  private String exchangeDlq;
+
   private AmqpConfiguration amqpConfiguration;
 
   public CreatedConfiguration(
```
jochen-vermeulen: There is a high degree of similarity between all configurations in this package. Is there anything you can do to generalize?

src/main/java/eu/europeana/metis/sandbox/domain/Event.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  public Event(T body, Step step, String exception) {
+    requireNonNull(body, "Body must not be null");
+    requireNonNull(step, "Step must not be null");
+    this.status = exception == null ? Status.SUCCESS : Status.FAIL;
+    this.body = body;
+    this.exception = exception;
+    this.step = step;
+  }
```
jochen-vermeulen: It's better to let two of these constructors call the third directly, instead of having three blocks that do more or less the same.

src/main/java/eu/europeana/metis/sandbox/domain/Record.java
```diff
@@ -16,33 +16,26 @@
   private final String datasetName;
```
jochen-vermeulen: This class has unused imports

src/main/java/eu/europeana/metis/sandbox/entity/RecordLogEntity.java
```diff
@@ -24,13 +24,14 @@
   @Enumerated(EnumType.STRING)
   private Status result;
 
-  @Column
+  @Column(columnDefinition = "VARCHAR(MAX)")
```
jochen-vermeulen: I think this column definition is very implementation dependent (i.e. not all databases support this or interpret it the same way). 
You should probably choose a sensible limit and truncate the error message here so that if it is bigger than your limit you don't create an error (which it currently does).

Note: this declaration also occurs above where the content is stored. I think this is often bigger than whatever 'MAX' is, and should probably be stored using a TEXT or even BLOB style column.

src/main/java/eu/europeana/metis/sandbox/service/dataset/AsyncDatasetPublishServiceImpl.java
```diff
@@ -3,8 +3,11 @@
 import static com.google.common.base.Preconditions.checkArgument;
 import static java.util.Objects.requireNonNull;
 
+import eu.europeana.metis.sandbox.common.Step;
+import eu.europeana.metis.sandbox.common.exception.RecordProcessingException;
 import eu.europeana.metis.sandbox.common.exception.ServiceException;
```
jochen-vermeulen: More unused imports.

src/main/java/eu/europeana/metis/sandbox/service/record/RecordLogServiceImpl.java
```diff
@@ -24,21 +25,23 @@ public RecordLogServiceImpl(RecordLogRepository repository) {
   }
 
   @Override
-  public void logRecord(Record record) {
-    requireNonNull(record, "Record must not be null");
+  public void logRecordEvent(Event<Record> recordEvent) {
```
jochen-vermeulen: The log object in this class is misnamed and not used.

src/main/java/eu/europeana/metis/sandbox/service/workflow/ExternalValidationService.java
```diff
@@ -1,8 +1,16 @@
...
 
+import eu.europeana.metis.sandbox.common.exception.RecordProcessingException;
 import eu.europeana.metis.sandbox.domain.Record;
 
 public interface ExternalValidationService {
 
+  /**
+   *
+   * @param record
+   * @return
```
jochen-vermeulen: Finish the javadoc :)

src/main/java/eu/europeana/metis/sandbox/service/workflow/ExternalValidationServiceImpl.java
```diff
@@ -1,14 +1,30 @@
...
   public Record validate(Record record) {
-    return Record.from(record, Step.VALIDATE_EXTERNAL);
+    String recordOrdered;
+    try {
+      recordOrdered = orderingService.performOrdering(record.getContent());
+    } catch (TransformationException e) {
+      throw new RecordProcessingException(record.getRecordId(), e);
+    }
+
+    return Record.from(record, recordOrdered);
```
jochen-vermeulen: I assume that validation is to be inserted here at a later stage?

src/main/java/eu/europeana/metis/sandbox/service/workflow/OrderingService.java
```diff
@@ -1,8 +1,17 @@
 package eu.europeana.metis.sandbox.service.workflow;
 
-import eu.europeana.metis.sandbox.domain.Record;
+import eu.europeana.metis.sandbox.common.exception.NonRecoverableServiceException;
```
jochen-vermeulen: Unused import!
Luckily my IDE tells me... :)

src/main/java/eu/europeana/metis/sandbox/service/workflow/OrderingServiceImpl.java
```diff
@@ -1,13 +1,33 @@
 package eu.europeana.metis.sandbox.service.workflow;
```
jochen-vermeulen: And another unused import... this is a good way to get many review comments!

src/main/java/eu/europeana/metis/sandbox/common/amqp/RecordMessageConverter.java
acerdas84: True, it was included because the interface has it and the IDE inherited when doing the override, I will remove it

src/main/java/eu/europeana/metis/sandbox/domain/Event.java
acerdas84: Not really, I was experimenting, but you are right is not going to be anything else than a Record, I will change it

src/main/java/eu/europeana/metis/sandbox/domain/Event.java
acerdas84: Yes I thought of that, but resolution was not working, anyways I already change this code in the next review I have pending to send

src/main/java/eu/europeana/metis/sandbox/entity/RecordLogEntity.java
acerdas84: columnDefinition affects DDL generation, I have it for the in memory database when running the integration tests. But persistent postgres is not generating the database from this definition, the table itself must exist or the application will fail, and the field error in the DB can be something else like TEXT or other type. But I see it can be confusing, I move the definition of this type to a custom script for the integration test

src/main/java/eu/europeana/metis/sandbox/service/record/RecordLogServiceImpl.java
acerdas84: Is not logging to a log file itself, but what I meant is to log every event in the database, that is where the name logRecordEvent comes from, and that is what I am saving/logging, an event into the "event log". Anyways I could name it saveRecordEvent, but is just a matter of taste from my point of view and I have no strong feelings for it, so let me know what you prefer

src/main/java/eu/europeana/metis/sandbox/service/workflow/ExternalValidationService.java
acerdas84: I have it for the next review about validation, but will add it here

src/main/java/eu/europeana/metis/sandbox/service/workflow/ExternalValidationServiceImpl.java
acerdas84: yes

src/main/java/eu/europeana/metis/sandbox/service/workflow/OrderingService.java
acerdas84: yeah I see I missed a lot of those ones, I will be more careful

src/main/java/eu/europeana/metis/sandbox/config/amqp/CreatedConfiguration.java
acerdas84: I found [this](https://docs.spring.io/spring-amqp/reference/html/#collection-declaration) I think is the best option available, let me know what do you think so I can look in more depth

src/main/java/eu/europeana/metis/sandbox/service/record/RecordLogServiceImpl.java
acerdas84: Sorry I realized what you were talking about, I just removed it

src/main/java/eu/europeana/metis/sandbox/config/amqp/CreatedConfiguration.java
acerdas84: I will explore this in the next review, so that we can close this one and I can merge it to the other one I have pending if that is fine with you

src/main/java/eu/europeana/metis/sandbox/config/SandboxConfig.java
```diff
@@ -24,12 +31,30 @@
...
+  private String edmSorterUrl = null;
+
   @Bean
   @Scope("prototype")
   XPathFactory xPathFactory() {
     return XPathFactory.newDefaultInstance();
   }
 
+  @Bean
+  @Scope("prototype")
```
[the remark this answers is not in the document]
acerdas84: I've seen both styles, seems that both are valid as stated [here](https://github.com/spring-projects/spring-framework/issues/19905), so it is a matter of taste I guess. If using the constant is preferable I will switch. Let me know your thoughts

src/main/java/eu/europeana/metis/sandbox/service/workflow/OrderingServiceImpl.java
```diff
@@ -1,13 +1,33 @@
...
+
+  public OrderingServiceImpl(
+      ObjectFactory<XsltTransformer> orderObjectFactory) {
+    this.orderObjectFactory = orderObjectFactory;
+  }
+
   @Override
-  public Record performOrdering(Record record) {
-    return record;
+  public String performOrdering(String record) throws TransformationException {
```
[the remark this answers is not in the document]
acerdas84: At first I grab the idea of strings from the data-checker, (but that does not do the ordering step) which treats records as a string and libraries it uses were handling it. For example for the validation step I am using `ValidationExecutionService.singleValidation` method which takes a String for the validation. 

Maybe I am not sure how many metis libraries need to change (or how hard are those changes) to make it work with a byte array. Can we revisit this topic in the Normalization step story? I think by that moment I will have a better idea of how much I am doing this transformation back and forth. And also if it makes sense, then change it to byte array should be easy (in sandbox)

src/main/java/eu/europeana/metis/sandbox/config/SandboxConfig.java
```diff
@@ -24,12 +31,30 @@
...
+
+  private String edmSorterUrl() throws IOException {
+    if(edmSorterUrl == null) {
+      edmSorterUrl = resourceLoader.getResource("classpath:edm/edm.xsd.sorter.xsl").getURL().toString();
+    }
+    return edmSorterUrl;
+  }
+
   @Bean
   Executor taskExecutor() {
```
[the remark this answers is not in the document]
acerdas84: I renamed that one to asyncDatasetPublishServiceTaskExecutor, and is used by the AsyncDatasetPublishServiceImpl for the Completable future to use.
And for the queues, each queue has its own RabbitListenerContainerFactory with 5 threads to start and a max of 10, but that is setup in each queue config

src/main/java/eu/europeana/metis/sandbox/config/SandboxConfig.java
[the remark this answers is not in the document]
jochen-vermeulen: I would say to use the constant: that way there is immediate access to the documentation as well, so we know why this particular scope was used here.

src/main/java/eu/europeana/metis/sandbox/config/SandboxConfig.java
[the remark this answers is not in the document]
jochen-vermeulen: OK, I added a note to the deployment ticket to make sure that this is all set up properly.

src/main/java/eu/europeana/metis/sandbox/service/workflow/OrderingServiceImpl.java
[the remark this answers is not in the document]
jochen-vermeulen: OK. Please be sure to add a note to that ticket so that we don't forget it.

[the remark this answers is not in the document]
acerdas84: Added note to https://europeana.atlassian.net/browse/MET-2543

PR #3, opened 2020-03-25
Feat/met 2540 validate xml records

acerdas84:
Validation changes:
eu/europeana/metis/sandbox/config/SandboxConfig.java
eu/europeana/metis/sandbox/service/workflow/ExternalValidationServiceImpl.java
predefined-schemas.properties.example

Comments from MET-2539
1. High degree of similarity between all configurations in this package. Aim to generalize:
eu/europeana/metis/sandbox/config/amqp/AmqpConfiguration.java
eu/europeana/metis/sandbox/config/amqp/WorkflowListenerConfig.java
2. Three very similar constructors, two of which should directly call the third: eu/europeana/metis/sandbox/domain/Event.java


src/main/java/eu/europeana/metis/sandbox/common/amqp/RecordMessageConverter.java
```diff
@@ -36,6 +46,12 @@ public Message toMessage(Object object, MessageProperties messageProperties) {
 
     Event recordEvent = (Event) object;
     Record record = recordEvent.getBody();
+    String recordException;
+    try {
+      recordException = recordEvent.getException() != null? objectMapper.writeValueAsString(recordEvent.getException()) : null;
```
jochen-vermeulen: Why do we need to do this? `recordEvent.getException()` is already a string. So we don't need to convert this, do we?

src/main/java/eu/europeana/metis/sandbox/common/amqp/RecordMessageConverter.java
```diff
@@ -64,8 +80,13 @@ public Object fromMessage(Message message) {
     String country = properties.getHeader(COUNTRY);
     String step = properties.getHeader(STEP);
     String content = new String(message.getBody(), DEFAULT_CHARSET);
-    LongString exceptionObject = properties.getHeader(EXCEPTION);
-    String exception = exceptionObject == null ? null : exceptionObject.toString();
+    LongString exceptionString = properties.getHeader(EXCEPTION);
+    Exception exception;
+    try {
+      exception = exceptionString == null ? null : objectMapper.readValue(exceptionString.toString(), Exception.class);
```
jochen-vermeulen: Same as above.

src/main/java/eu/europeana/metis/sandbox/config/SandboxConfig.java
```diff
@@ -76,4 +88,32 @@ String initialQueue() {
...
+  SchemaProvider schemaProvider() throws IOException {
+    return new SchemaProvider(PredefinedSchemasGenerator.generate(schemaProperties()));
+  }
+
+  private Properties schemaProperties() throws IOException {
+    PropertiesFactoryBean propertiesFactoryBean = new PropertiesFactoryBean();
+    propertiesFactoryBean.setLocation(new ClassPathResource("predefined-schemas.properties"));
+    propertiesFactoryBean.afterPropertiesSet();
+    return propertiesFactoryBean.getObject();
+  }
```
jochen-vermeulen: I guess this may be impacted by another remark I made. The question is: how do we handle workflow step-specific configuration like the validation schema information? We don't want separate property files for all steps, do we? Is it a good idea to integrate that into the application.yml? And if this implies a change in the validation library in Metis, that's fine.

src/main/java/eu/europeana/metis/sandbox/config/amqp/WorkflowListenerConfig.java
```diff
@@ -0,0 +1,84 @@
...
+      ConnectionFactory connectionFactory) {
+    return getSimpleRabbitListenerContainerFactory(configurer, connectionFactory);
+  }
+
+  @Bean
+  SimpleRabbitListenerContainerFactory indexedFactory(
+      SimpleRabbitListenerContainerFactoryConfigurer configurer,
+      ConnectionFactory connectionFactory) {
+    return getSimpleRabbitListenerContainerFactory(configurer, connectionFactory);
+  }
```
jochen-vermeulen: This is 8 times exactly the same method... is there nothing we can do here?

src/main/java/eu/europeana/metis/sandbox/service/validation/ValidationServiceConfigImpl.java
```diff
@@ -0,0 +1,17 @@
+package eu.europeana.metis.sandbox.service.validation;
+
+import eu.europeana.validation.service.ValidationServiceConfig;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Service;
+
+@Service
+class ValidationServiceConfigImpl implements ValidationServiceConfig {
```
jochen-vermeulen: Perhaps this is a place where more validation configuration options could be added (see other remarks)?

src/main/java/eu/europeana/metis/sandbox/service/workflow/ExternalValidationService.java
```diff
@@ -12,6 +13,7 @@
    * @return ordered record that is valid against EDM-EXTERNAL schema
    * @throws NullPointerException      if record is null
    * @throws RecordProcessingException if records fails at ordering
+   * @throws RecordValidationException if records fails at validation
```
jochen-vermeulen: Should these exceptions ultimately inherit from RuntimeException? That sometimes makes it less obvious what is being thrown and what the calling methods should account for. Would it be a good idea to convert them to inherit from Exception instead?

src/main/java/eu/europeana/metis/sandbox/service/workflow/ExternalValidationServiceImpl.java
```diff
@@ -1,29 +1,46 @@
...
 import eu.europeana.metis.sandbox.domain.Record;
 import eu.europeana.metis.transformation.service.TransformationException;
+import eu.europeana.validation.service.ValidationExecutionService;
 import org.springframework.stereotype.Service;
 
 @Service
 class ExternalValidationServiceImpl implements ExternalValidationService {
 
-  private OrderingService orderingService;
+  private final static String SCHEMA = "EDM-EXTERNAL";
```
jochen-vermeulen: "SonarLint: Reorder the modifiers to comply with the Java Language Specification"

src/test/java/eu/europeana/metis/sandbox/SandboxApplicationTests.java
```diff
@@ -6,8 +6,8 @@
 @SpringBootTest
 class SandboxApplicationTests {
 
-	@Test
-	void contextLoads() {
-	}
+  @Test
+  void contextLoads() {
+  }
```
jochen-vermeulen: Is this test supposed to be empty?

src/main/java/eu/europeana/metis/sandbox/config/amqp/WorkflowListenerConfig.java
acerdas84: Each bean is for a different queue (and there is one queue per workflow step), code looks the same because right now all of this beans are using the same default config that I specified in the application.yml, like the individual threads they are using to consume from the queues. So while they all look the same they are defined for different queues, and the requirements of those queues might make the config for each of these beans to change (needing more consumers, or changing the retry policies). But I did encapsulate the repetitive code in the last method `getSimpleRabbitListenerContainerFactory`

src/main/java/eu/europeana/metis/sandbox/service/validation/ValidationServiceConfigImpl.java
acerdas84: This one is a `@Service` (but yes the name does not help) it was a requirement of the `ValidationExecutionService` that is in metis-validation-service. (I followed the approach used in metis-validation-rest which injects `ValidationExecutionService` in `ValidationController`)

I am using `ValidationExecutionService` in the `ExternalValidationServiceImpl` but to be able to inject that service I needed to provide the beans that it depends on, which are
`ValidationServiceConfig validationServiceConfig` which is this implementation
`ClasspathResourceResolver lsResourceResolver` bean defined in configuration file
`SchemaProvider schemaProvider` bean defined in configuration file

src/main/java/eu/europeana/metis/sandbox/service/workflow/ExternalValidationService.java
acerdas84: Probably you already notice all exceptions I created inherit from `RuntimeException`, I think there is debate on the use of checked exceptions and based on my experience I lean towards using unchecked exceptions but documenting them in the interface contract so that whoever uses it will know what to account for (as Joshua Bloch recommends to document every exception your method can throw) and handle or not if they want. I am not saying my approach is right or wrong, I had this talk on the past with other colleagues and listen and share thoughts on the subject and most of the time I go for an unchecked exceptions hierarchy, and handle them as convenient. But also I took some time to read again about the arguments and collected some interesting links on the topic that you can check at the bottom of this comment.

From my experience with checked exceptions I notice most of the time there is not much you can do when they happen, you mostly re-throw or catch them, and when you do catch them is to log or to wrap them into a higher level exception to re-throw again, most of the time you let them go up until they are logged, and the difference between checked and unchecked becomes that you have to declare them in your method declaration or not. So the approach that I take in general is to create an unchecked exceptions hierarchy from the project, then use those in the service level (wrapping other low level exceptions with the service ones) and document the Interface of each service to declare every exception it throws, then for example in Sandbox there are 2 clients/users of the service layer, the controller and consumer layers, which know and understand the expected exceptions from the documentation in the services interface and they handle or let them go up as they think is appropriate.

Again not saying is right or wrong, just my own experience, and since is an interesting topic for me I took time to collect info and try to make a point (and for sure there could be a lot of counter arguments). 
I just wanted to share some thoughts on the subject, but also I think there could be good reasons to use checked exceptions like most [recommendations](https://docs.oracle.com/javase/tutorial/essential/exceptions/runtime.html) states 

> If a client can reasonably be expected to recover from an exception, make it a checked exception

But for me to find that point is what is not that clear, like in this case `RecordValidationException` is thrown because the record itself is not valid, so how will the user of the API (the consumer layer) can recover from that. Also quoting [Effective Java](https://www.oreilly.com/library/view/effective-java-3rd/9780134686097/)

> This burden may be justified if the exceptional condition cannot be prevented by proper use of the API and the programmer using the API can take some useful action once confronted with the exception. Unless both of these conditions are met, an unchecked exception is appropriate

Which is like most of the time for me. But I am not against and for sure I will change it if the standards you have advocate for their use, I just wanted to share my own experience and see what do you think about the topic

Here are some links for the thoughts on the topic:

- https://testing.googleblog.com/2009/09/checked-exceptions-i-love-you-but-you.html
- https://www.javaworld.com/article/3142626/are-checked-exceptions-good-or-bad.html
- https://stackoverflow.com/questions/6115896/understanding-checked-vs-unchecked-exceptions-in-java
- https://stackoverflow.com/questions/4609870/why-hibernate-changed-hibernateexception-to-unchecked-runtimeexception
- https://docs.oracle.com/javase/tutorial/essential/exceptions/runtime.html
- https://stackoverflow.com/questions/613954/the-case-against-checked-exceptions
- http://literatejava.com/exceptions/checked-exceptions-javas-biggest-mistake/
- https://www.artima.com/intv/solid.html
- http://tutorials.jenkov.com/java-exception-handling/checked-or-unchecked-exceptions.html
- https://stackoverflow.com/questions/23178952/why-spring-handles-only-unchecked-exceptions/23179413

src/main/resources/application.yml
```diff
@@ -57,6 +57,9 @@ sandbox:
         core-pool-size: 10
         max-pool-size: 40
         thread-prefix: sandbox-
+  validation:
+    executor:
+      pool-size: 10
   rabbitmq:
     exchange:
       name: sandbox
```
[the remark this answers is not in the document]
acerdas84: I debugged the code and it works both ways, but [spring docs](https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-application-properties.html#integration-properties) state what you suggested. 

Just out of curiosity, my IDE does not tells me anything about it, are you using a plugin or something that validates that?

src/test/java/eu/europeana/metis/sandbox/SandboxApplicationTests.java
acerdas84: This one is added by the initialzr when creating the project, its purpose its to be an Integration test to validate that the application context loads successfully, which in practice is more useful than it sounds

src/main/resources/predefined-schemas.properties.example
```diff
@@ -0,0 +1,9 @@
+predefinedSchemas=EDM-INTERNAL,EDM-EXTERNAL
+
+predefinedSchemas.EDM-INTERNAL.url=
+predefinedSchemas.EDM-INTERNAL.rootLocation=
+predefinedSchemas.EDM-INTERNAL.schematronLocation=
+
+predefinedSchemas.EDM-EXTERNAL.url=
+predefinedSchemas.EDM-EXTERNAL.rootLocation=
+predefinedSchemas.EDM-EXTERNAL.schematronLocation=
```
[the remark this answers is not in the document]
acerdas84: Actually I found a way to set it in the yml file (first time I was not able to do it), I will apply it and you can check for feedback

src/main/java/eu/europeana/metis/sandbox/config/SandboxConfig.java
acerdas84: I found a way to keep the config in the yml file, at first the reason I had to separate was that I did not found a way to have all together and provide the `Properties` to the `PredefinedSchemasGenerator.generate` method.
I will upload the way I found for you to check on it

src/main/java/eu/europeana/metis/sandbox/common/amqp/RecordMessageConverter.java
acerdas84: I think there was a merge issue when I merged developed into my branch, I will fix it and then it will make sense

src/main/java/eu/europeana/metis/sandbox/common/amqp/RecordMessageConverter.java
acerdas84: same here

src/main/java/eu/europeana/metis/sandbox/config/SandboxConfig.java
```diff
@@ -76,4 +88,32 @@ String initialQueue() {
...
+
+  @Bean
+  SchemaProvider schemaProvider() {
+    return new SchemaProvider(PredefinedSchemasGenerator.generate(schemaProperties()));
+  }
+
+  private Properties schemaProperties() {
+    YamlPropertiesFactoryBean propertiesFactoryBean = new YamlPropertiesFactoryBean();
+    propertiesFactoryBean.setResources(new ClassPathResource("application.yml"));
```
acerdas84, line 104: Now instead of using a properties file I am using the application.yml, and I am able to get the `PredefinedSchemasGenerator.generate` the properties it needs

src/test/resources/schema.sql
```diff
@@ -1,18 +1,23 @@
+drop table if exists dataset;
```
acerdas84, line 1: This is the DB schema for integration testing

src/main/java/eu/europeana/metis/sandbox/common/amqp/RecordMessageConverter.java
```diff
@@ -36,6 +40,7 @@ public Message toMessage(Object object, MessageProperties messageProperties) {
 
     Event recordEvent = (Event) object;
     Record record = recordEvent.getBody();
+    Optional<EventError> eventError = recordEvent.getEventError();
```
acerdas84, line 43: I went away from using the Exception idea because at the end what I want to preserve is why the record event failed, so instead I created the EventError that will contain the error message and the stack trace of it, that I will be saving later. Now I removed the object mapper, and made a few changes to adjust the logic

src/main/java/eu/europeana/metis/sandbox/common/amqp/RecordMessageConverter.java
acerdas84: At the end I took another approach, but since this comment does not show anymore in the review I just want to make the statement here before resolving this one

src/main/resources/application.yml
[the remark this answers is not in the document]
jochen-vermeulen: No, I don't think I have a special plugin installed... I have IntelliJ and it just gave an error.

src/main/java/eu/europeana/metis/sandbox/service/workflow/ExternalValidationService.java
jochen-vermeulen: I will admit I have not checked out all the references you gave (yet... I may do so later). But thanks for writing all this, it has focused my thinking a bit.

I think that we are in agreement on an abstract level, meaning that unchecked exceptions should be reserved for 'stuff we can't do anything about' (i.e. we can't or shouldn't be expected to recover from). To me that sometimes translates into whether we can expect an exception to occur in the normal flow of things, or whether the user (whoever that is) should see a 'nice' message or just a technical one.

The example you gave is in fact an interesting one. The RecordValidationException currently only occurs when you are checking record validity. This is in my opinion something that the tool should expect to happen (input records are often invalid, and is in fact something that people are going to use the sandbox for, to check this). So in line with our shared understanding, I would argue that it SHOULD be a checked exception, forcing code higher-up to deal with it properly and providing a meaningful feedback to the user.

However, perhaps in a different context (e.g. transformation) we know we have validated the record and the RecordValidationException should no longer occur, if it does there is a software problem that the tool is NOT expected to recover from. So perhaps there it should be thrown as or encapsulated in an unchecked/runtime exception.

So in short, I think we should spend some time identifying which exceptions are 'expected' (i.e. the user's or input's 'fault') and which aren't (i.e. the system's or programmer's fault). And make those in the former category checked? What do you think?

src/main/java/eu/europeana/metis/sandbox/service/validation/ValidationServiceConfigImpl.java
jochen-vermeulen: Well, it seems to be a class that reads properties (in fact only one property) from the configuration file and return it. So I thought that this would be a place to add similar code for the properties that the validation code needs and pass that along explicitly. But it seems you found a way to do it implicitly. I will remark further on it elsewhere.

src/test/java/eu/europeana/metis/sandbox/SandboxApplicationTests.java
jochen-vermeulen: As long as it does something useful :)
Perhaps you could include a comment to that effect in the code?

src/main/resources/application.yml
```diff
@@ -98,3 +98,11 @@ sandbox:
...
           queue: sandbox.record.indexed
           dlq: ${sandbox.rabbitmq.queues.record.indexed.queue}.dlq
+
+predefinedSchemas: EDM-INTERNAL,EDM-EXTERNAL
+predefinedSchemas.EDM-INTERNAL.url:
+predefinedSchemas.EDM-INTERNAL.rootLocation:
+predefinedSchemas.EDM-INTERNAL.schematronLocation:
+predefinedSchemas.EDM-EXTERNAL.url:
+predefinedSchemas.EDM-EXTERNAL.rootLocation:
+predefinedSchemas.EDM-EXTERNAL.schematronLocation:
```
jochen-vermeulen, line 108: In true YAML fashion, could this be reworded as follows?
```
predefinedSchemas: EDM-INTERNAL,EDM-EXTERNAL
  EDM-INTERNAL:
    url:
    rootLocation:
    schematronLocation:
  EDM-EXTERNAL:
    url:
    rootLocation:
    schematronLocation:
```
Or would that not work?

jochen-vermeulen: Also: From the point of view of the configuration file, would it not make more sense to include it in the validation specific section at sandbox.validation? More configuration will/may be needed for other components, and perhaps it would be good to structure this properly. Even if this has some effect in the code, this may be worth it. What do you think?

acerdas84: Yes I actually try that, but it does not work because of the nature of the properties
```
predefinedSchemas: EDM-INTERNAL,EDM-EXTERNAL
  EDM-INTERNAL:
```
You cannot do a property with a value and then a hierarchy, YAML will not work, I look around and this one was the way to go when having this kind of hierarchy. On the other hand I could reorder to something like
```
sandbox:
  validation:
    predefinedSchemas: 
      schemas: EDM-INTERNAL,EDM-EXTERNAL
      EDM-INTERNAL:
        url:
        rootLocation:
        schematronLocation:
      EDM-EXTERNAL:
        url:
        rootLocation:
        schematronLocation:
```
And then just adjust it when using the method I need.
I will make a note for MET-2541 to look into this

src/main/java/eu/europeana/metis/sandbox/service/workflow/ExternalValidationService.java
acerdas84: Sounds good, but I think the evaluation itself deserves its own story, I can create a ticket in the backlog to be tackled on its own, for us to define the Exceptions hierarchy and explore which ones are checked and which not. (Also is the first time I think of them in terms of the kind of message that you can retrieve to the user, that is a interesting point)

src/main/resources/application.yml
acerdas84: [MET-2541](https://europeana.atlassian.net/secure/RapidBoard.jspa?rapidView=61&modal=detail&selectedIssue=MET-2541)

src/main/java/eu/europeana/metis/sandbox/service/workflow/ExternalValidationService.java
acerdas84: Created [MET-2622](https://europeana.atlassian.net/browse/MET-2622)

src/test/java/eu/europeana/metis/sandbox/SandboxApplicationTests.java
acerdas84: Sure I will do that now

src/main/resources/application.yml
jochen-vermeulen: Yes, I was afraid of that... your suggested structure looks like an improvement. We may not actually need the 'schemas' line: it is implied by the two children that its parent has.

PR #4, opened 2020-04-03
Feat/met 2541 transform xml records

acerdas84:
Transformation logic

From MET-2540: revisit predefinedSchemas properties to arrange them in a more YAML compatible way and adjust the hierarchy to sandbox.validation


src/main/java/eu/europeana/metis/sandbox/common/amqp/RecordMessageConverter.java
```diff
@@ -20,17 +20,17 @@
...
-  private static final String STACK_TRACE = "stackTrace";
+  protected static final String RECORD_ID = "recordId";
+  protected static final String DATASET_ID = "datasetId";
+  protected static final String DATASET_NAME = "datasetName";
+  protected static final String LANGUAGE = "language";
+  protected static final String COUNTRY = "country";
+  protected static final String STATUS = "status";
+  protected static final String STEP = "step";
+  protected static final String ERROR = "error";
+  protected static final String STACK_TRACE = "stackTrace";
```
acerdas84, line 31: Set as protected for testing purposes

src/main/java/eu/europeana/metis/sandbox/service/workflow/TransformationServiceImpl.java
```diff
@@ -1,13 +1,48 @@
...
   public Record transform(Record record) {
-    return record;
+    requireNonNull(record, "Record must not be null");
+
+    String recordTransformed;
+    try {
+      var europeanaGeneratedIdsMap = new EuropeanaIdCreator()
+          .constructEuropeanaId(record.getContent(), record.getDatasetId());
+      var xsltTransformer = getTransformer(record.getDatasetName(),
+          record.getCountry().xmlValue(), record.getLanguage().xmlValue());
```
jochen-vermeulen, line 34: The variable you're declaring here already exists as a class field. It's not good practice.

src/main/java/eu/europeana/metis/sandbox/config/SandboxConfig.java
```diff
@@ -67,6 +69,21 @@ private String edmSorterUrl() throws IOException {
...
+  @Bean
+  @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
+  XsltTransformer xsltTransformer(String datasetName, String edmCountry, String edmLanguage)
+      throws IOException, TransformationException {
+    return new XsltTransformer(defaultXsltTransformerUrl(), datasetName, edmCountry, edmLanguage);
+  }
+
+  private String defaultXsltTransformerUrl() throws IOException {
+    if (defaultXsltUrl == null) {
+      defaultXsltUrl = resourceLoader.getResource("classpath:edm/default.xslt.xsl").getURL()
```
jochen-vermeulen, line 81: You are including the default xslt as a file in this repository. Wouldn't it be preferable to be more flexible? Just like we can upload a new xslt in Metis at any time, we should allow the xslt to be changed at runtime here as well. Perhaps have a line in the config file with a file location? Or perhaps some other solution? @stzanakis , you have any ideas?

stzanakis: Because this is an "open" service and we do not have authentication, I would opt out of having an upload of a default xslt.
That being said though, I agree we should have it configurable and change it(if required).
To do so it can either be a configuration property pointing to a path on the file system but that implies that the default xslt will be packaged with the .war file(more likely from jenkins) and a new xslt will require a new build. That is because of bluemix deployments limitations.
Or it has to be part of the database and loaded from there on a restart or a trigger/reload call.

p.s. generally the default xslt changes rarely so we can opt in on any of the options.

acerdas84: is there a centralized place where you pull it ? a Url that I can use? Or how is it managed today ?
I could remove it from the repository and it can be injected by the deployment like the properties are, so that is not in the repo and it can change separately

stzanakis: In metis after the first deployment of metis-core there is an endpoint which we use to upload the xslt and it is inserted in the db. But there we have authentication/authorization in place so it's fine to do so.
The metis-core also supplies another endpoint to GET the latest default xslt without using authentication, you could use that but that required the service to be up and running.
Perhaps locally handling it is better..

acerdas84: In that case I think I like the option of providing the file during build/deployment like the application.yml, if that is possible, since the story says 
> EDM records are transformed against the default XSLT transformation used by Metis. The user has no option for uploading his own XSLT
And because the default does not change that often I think the ability to switch by deploying will be good enough for now. 
Tradeoff of handling locally is that we need to remember that the sandbox has to be deployed in order to apply changes.
To summarize, proposal is to remove the xstl from the repo, and let it be provided by jenkins build

src/main/java/eu/europeana/metis/sandbox/config/SandboxConfig.java
```diff
@@ -67,6 +70,13 @@ private String edmSorterUrl() throws IOException {
     return edmSorterUrl;
   }
 
+  @Bean
+  @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
+  XsltTransformer xsltTransformer(String datasetName, String edmCountry, String edmLanguage)
+      throws TransformationException {
+    return new XsltTransformer(defaultXsltUrl, datasetName, edmCountry, edmLanguage);
```
acerdas84, line 77: now the `defaultXsltUrl` comes from a property in the application.yml

src/main/java/eu/europeana/metis/sandbox/config/SandboxConfig.java
```diff
@@ -97,15 +107,55 @@ SchemaProvider schemaProvider() {
...
     return new RecordMessageConverter();
   }
+
+  @Bean
+  @ConfigurationProperties(prefix = "sandbox.validation")
+  Schema schema() {
+    return new Schema();
+  }
+
+  @Value("${sandbox.transformation.xslt-url}")
```
acerdas84, line 121: Here is the property where the url will come, but it has to be a valid url string, like 
`file:/Users/user/documents/path/to/xstl/default.xsl`

src/main/resources/application.yml
```diff
@@ -60,6 +60,17 @@ sandbox:
...
+      EDM-INTERNAL:
+        url:
+        rootLocation:
+        schematronLocation:
+      EDM-EXTERNAL:
+        url:
+        rootLocation:
+        schematronLocation:
+  transformation:
+    xslt-url:
```
acerdas84, line 73: url goes here, it should be the url to the file which can be local or remote
Example: `file:/Users/user/documents/path/to/xstl/default.xsl`

PR #6, opened 2020-04-10
Feat/met 2543 normalize xml records

acerdas84:
Logic to normalize a record

Also:

- Applied [sonar](http://sonarqube.eanadev.org:9000/project/issues?id=eu.europeana.metis%3Ametis-sandbox&resolved=false&types=BUG) feedback 
- Document consumers
- Minor details

And last:
>Review comments from MET-2539: As a Sandbox user, I want my XML fields to be ordered in my xml records before validation to be solved in this ticket:
OrderingServiceImpl.java: I just realized that you're always looking at records as a string. Is it not better to store them, manage them and exchange them as byte array? This would avoid a lot of potential converting back and forth, like you're doing in this method. To me it seems the safest way to store them.

After doing more actions (external validation, transformation, internal validation, normalization and enrich) I saw almost all of those actions treat the record as a string, (only the transformation step has an input of byte array, and do not return a byte array, but a StringWriter)
So while treating the record as a byte array sounds a good idea, I feel it goes beyond this scope because the metis libraries are the ones relying on strings, and the only one actually getting a byte array do not even return that. For now I will leave the code as it is using string to represent the record


src/main/java/eu/europeana/metis/sandbox/common/exception/InvalidZipFileException.java
```diff
@@ -2,6 +2,8 @@
 
 public class InvalidZipFileException extends RuntimeException {
 
+  private static final long serialVersionUID = -2555540887797325483L;
```
acerdas84, line 5: I forgot about this until sonar remind me 😃

src/main/java/eu/europeana/metis/sandbox/config/SandboxConfig.java
```diff
@@ -137,7 +143,7 @@ private Properties schemaProperties() {
 
   private void addSchemaProperties(String key, Object value, Properties props) {
     if (value instanceof String) {
-      props.put(key, value);
+      props.setProperty(key, (String) value);
```
acerdas84, line 146: Sonarqube feedback

src/main/java/eu/europeana/metis/sandbox/domain/Record.java
```diff
@@ -74,11 +85,11 @@ public boolean equals(Object o) {
       return false;
     }
     Record record = (Record) o;
-    return recordId.equals(record.recordId) &&
+    return country == record.country &&
```
acerdas84, line 88: Sonarqube feedback

src/main/java/eu/europeana/metis/sandbox/consumer/workflow/TransformedConsumer.java
```diff
@@ -14,13 +14,18 @@
...
 import org.springframework.stereotype.Component;
 
+/**
+ * Consumes transformed events and performs internal validation to the contained record
+ * <br/>
+ * Publishes the result in the internally validated queue
+ */
 @Component
-public class TransformedConsumer {
+class TransformedConsumer {
```
jochen-vermeulen, line 23: All those consumers are starting to look alike... is there perhaps something we can do to generalize?

acerdas84: I did notice that about consumers, even attempt to do a generalization, but I failed with my first idea so I rolled back and need to think about it again, I will create a ticket to address this

acerdas84: https://europeana.atlassian.net/browse/MET-2663

PR #7, opened 2020-04-15
Feat/met 2544 enrich xml records

acerdas84:
Enrichment logic

Also: sonarqube feedback


src/main/java/eu/europeana/metis/sandbox/config/amqp/AmqpConfiguration.java
```diff
@@ -18,11 +19,6 @@
 
   private final MessageConverter messageConverter;
 
-  public AmqpConfiguration(
```
acerdas84, line 21: Moving constructor after field declarations

src/main/java/eu/europeana/metis/sandbox/config/amqp/AmqpConfiguration.java
```diff
@@ -98,55 +125,106 @@ AmqpTemplate amqpTemplate(ConnectionFactory connectionFactory) {
...
+        QueueBuilder.durable(dlqs.mediaProcessed).build(),
+        QueueBuilder.durable(dlqs.indexed).build()
     );
   }
 
   @Bean
   Declarables bindings() {
-    return getDeclarables(createdQueue, externalValidatedQueue, transformedQueue,
-        normalizedQueue, internalValidatedQueue, enrichedQueue, mediaProcessedQueue, indexedQueue);
+    return getDeclarables(queues);
```
acerdas84, line 155: Reduce the method parameters by using an object

src/main/java/eu/europeana/metis/sandbox/service/workflow/EnrichmentServiceImpl.java
```diff
@@ -1,13 +1,35 @@
...
+
   @Override
   public Record enrich(Record record) {
-    return record;
+    requireNonNull(record, "Record must not be null");
+
+    String result;
+    try {
+      result = enrichmentWorker.process(record.getContent());
+    } catch (DereferenceOrEnrichException | JiBXException | UnsupportedEncodingException e) {
```
acerdas84, line 29: I am catching all of these exceptions, but under the assupmtion that none of those will be a stopping point for the record, is that correct ?

src/main/java/eu/europeana/metis/sandbox/service/dataset/AsyncDatasetPublishServiceImpl.java
```diff
@@ -46,10 +46,8 @@ private void publish(Record record) {
     Event recordEvent = new Event(record, Step.CREATE);
     try {
       amqpTemplate.convertAndSend(initialQueue, recordEvent);
-    } catch (Exception e) {
-      log.error("There was an issue publishing the record: {} {}", record.getRecordId(),
-          e.getMessage(), e);
-      throw new RecordProcessingException(record.getRecordId(), e);
+    } catch (AmqpException e) {
+      LOGGER.error("There was an issue publishing the record: {} ", record.getRecordId(), e);
```
jochen-vermeulen, line 50: Just a question: You don't need to do anything with the error here? Is this not something an observer/user needs to know?

acerdas84: My reasoning was that this method is invoked in an asynchronous way, the caller method returns `CompletableFuture<Void>` and no one is using it to invoke `CompletableFuture.get()` (which is okay), so throwing the exception here will not be catched anywhere since it is running on its own thread, then I thought just logging it to have it in the logs will be enough.

But thinking it again, if someone at some point wants to invoke `CompletableFuture.get()` then throwing the exception could make sense.

At first I was logging because it is the place to do it, and throwing because there is nothing else I can do to fix things, then sonar gave me a warn of doing one of those but not both (I thought on ignoring it because of the scenario), then I pick logging no one is catching anything I throw here, but on the other hand it could be that at some point someone could want to know, so to summarize I think I will add the throw again, if that sounds good to you.

jochen-vermeulen: I thought about it a little bit, and I think we should look at it from a user's point of view. Whenever a record cannot advance from one queue to another, it ends up in a collection of errors that is later retrievable by the user. Perhaps we can do the same here? Send it to the same place that the other errors go so that the user knows what happened to his records? Then here we can just do the logging. Does that sound reasonable?

acerdas84: Yes, I will go that way

acerdas84: Moving this to [MET-2571](https://europeana.atlassian.net/browse/MET-2571)

PR #8, opened 2020-04-20
Feat/met 2545 media process xml records

acerdas84:
Process media from records

Also:

- Refactor to use record content as byte[] instead of String


src/main/java/eu/europeana/metis/sandbox/common/amqp/RecordMessageConverter.java
```diff
@@ -58,21 +54,21 @@ public Message toMessage(Object object, MessageProperties messageProperties) {
       properties.setHeader(STACK_TRACE, eventError.get().getStackTrace());
     }
 
-    return MessageBuilder.withBody(record.getContent().getBytes(DEFAULT_CHARSET))
+    return MessageBuilder.withBody(record.getContent())
```
acerdas84, line 57: Content is now a byte[]

src/main/java/eu/europeana/metis/sandbox/config/amqp/AmqpConfiguration.java
```diff
@@ -125,38 +98,56 @@ AmqpTemplate amqpTemplate(ConnectionFactory connectionFactory) {
   @Bean
   Declarables queues() {
     return new Declarables(
```
acerdas84, line 100: Sonarqube bug fix

src/main/java/eu/europeana/metis/sandbox/domain/Record.java
```diff
@@ -4,29 +4,40 @@
...
 
+  private static final Charset DEFAULT_CHARSET = StandardCharsets.UTF_8;
+
   private final String recordId;
   private final String datasetId;
   private final String datasetName;
   private final Country country;
   private final Language language;
-  private final String content;
+  private final byte[] content;
```
acerdas84, line 29: Now content is a `byte[]`. And while still preserve the String, it is created from the provided `byte[]`.
But now that this field is a `byte[]` it is no longer an immutable object, I thought of returning a copy of the `content` but then it will be expensive. So I decided to make a note on not modyfing the content byte array.
On the other hand it is now easier to pass the events through the message broker since the content is now a byte array, but we still need the String for some steps that take a String in the parameter

src/main/java/eu/europeana/metis/sandbox/domain/Record.java
```diff
@@ -38,6 +48,19 @@ private Record(String recordId, String datasetId, String datasetName,
    * @throws NullPointerException if any parameter is null
    */
   public static Record from(Record record, String content) {
+    requireNonNull(content);
+    return from(record, content.getBytes(DEFAULT_CHARSET));
```
acerdas84, line 52: There are steps (metis libraries) that return a String as the result of the process. So we keep this one, and is preferred since it will use UTF-8 charset

src/main/java/eu/europeana/metis/sandbox/domain/Record.java
```diff
@@ -75,10 +98,25 @@ public Language getLanguage() {
...
+   * @implNote Overwriting this field contents after construction could cause problems.
+   * <br />
+   * We are not making a copy of it because it
+   * is expensive and Record object is expected to be use as a non mutable object
+   */
+  public byte[] getContent() {
     return this.content;
   }
 
+  public String getContentString() {
```
acerdas84, line 113: Lazy initialization, only if the content is required as a String it will be populated

src/main/java/eu/europeana/metis/sandbox/util/ZipFileReader.java
```diff
@@ -0,0 +1,78 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class provides the functionality of reading zip files.
+ * <br /><br />
+ * TODO search ZipFileReader class in project metis-common and add a method that also returns
+ *   records as List<byte[]> then replace this class with that one
+ */
+public class ZipFileReader {
```
acerdas84, line 22: This is a copy of a class in metis-common project, but it returns a list of byte array instead of a list of strings. I think a method that returns this could be added there and then we can replace this class, but for now I modify it and put it here

src/main/java/eu/europeana/metis/sandbox/domain/Record.java
jochen-vermeulen: Please let us know which steps require a string! Perhaps internally it is converted to a byte array again and it would be best to expose a method that accepts a byte array.

src/main/java/eu/europeana/metis/sandbox/util/ZipFileReader.java
jochen-vermeulen: Let's do this the proper way. Let's create that method in metis common!

src/main/java/eu/europeana/metis/sandbox/util/ZipFileReader.java
```diff
@@ -0,0 +1,78 @@
...
+        if (reject(zipEntry)) {
+          continue;
+        }
+        result.add(IOUtils.toByteArray(zipFile.getInputStream(zipEntry)));
+      }
+      return result;
+    }
+  }
+
+  boolean reject(ZipEntry zipEntry) {
```
jochen-vermeulen, line 74: Apparently, access here can be private.

src/test/java/eu/europeana/metis/sandbox/service/workflow/MediaProcessingServiceImplTest.java
```diff
@@ -1,10 +1,357 @@
...
+    Record result = service.processMedia(record);
+
+    assertArrayEquals(content, result.getContent());
+    verify(thumbnailStoreService).store(List.of());
+  }
+
+  @Test
+  void processMedia_storeThumbnails_expectFail()
+      throws RdfConverterException, RdfDeserializationException, MediaProcessorException,
+      MediaExtractionException, IOException {
```
jochen-vermeulen, line 260: Apparently, IOException is not thrown here.

src/main/java/eu/europeana/metis/sandbox/service/workflow/MediaProcessingServiceImpl.java
```diff
@@ -1,13 +1,162 @@
...
+  private List<RdfResourceEntry> getRdfResourceEntries(Record record, byte[] content,
+      RdfDeserializer rdfDeserializer) {
+    List<RdfResourceEntry> resourceEntries;
+    try {
+      resourceEntries = rdfDeserializer
+          .getResourceEntriesForMediaExtraction(content);
+    } catch (RdfDeserializationException e) {
+      throw new RecordProcessingException(record.getRecordId(), e);
+    }
+    return resourceEntries;
```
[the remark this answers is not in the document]
acerdas84: Yes, I will do that

src/main/java/eu/europeana/metis/sandbox/service/workflow/MediaProcessingServiceImpl.java
```diff
@@ -1,13 +1,162 @@
...
+    requireNonNull(record, "Record must not be null");
+
+    var inputRdf = record.getContent();
+    var rdfDeserializer = getRdfDeserializer(record);
+
+    // Get resource entries
+    var resourceEntries = getRdfResourceEntries(record, inputRdf, rdfDeserializer);
+
+    // Perform media processing
+    var resourceExtractionResults = getResourceExtractionResults(record, resourceEntries);
```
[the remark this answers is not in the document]
acerdas84: I will explore going that path

src/main/java/eu/europeana/metis/sandbox/service/workflow/OrderingService.java
```diff
@@ -12,5 +12,5 @@
    * @throws NullPointerException    if xml is null
    * @throws TransformationException if ordering fails
    */
-  String performOrdering(String xml) throws TransformationException;
+  String performOrdering(byte[] xml) throws TransformationException;
```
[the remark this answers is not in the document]
acerdas84: Well is tricky because  `XsltTransformer.transform(byte[])` returns a `StringWriter` which I will probably still do a `toString().getBytes(Charset)` to get the byte[] as a return, which then will need to become a String again because validation library takes a String as a parameter.
But I am up to change all of this when libraries support receiving and returning a byte[] as the record representation, but I have the feeling that will take more work than it sounds

src/main/java/eu/europeana/metis/sandbox/util/ZipFileReader.java
acerdas84: Yes I thought of that, but then I will have to require snapshot versions of metis libraries instead of current release, which I could do, but I thought it was better for the sandbox to stay as a client of metis libraries as they are now (release 1.8.0 I think).

I see you evolve metis libraries in sync with ecloud code; so probably in the future sandbox releases could be in sync with metis complete environment releases. For the first production deployment I did not think of depending on metis deployment but maybe that is wrong.

But if we are not making the code available in sync then even if I go and add the method to metis-common I will only be able to use it if looking at a snapshot.

I am open to suggestions. Maybe I am not understanding something on the process of release.
What do you think?

src/main/java/eu/europeana/metis/sandbox/domain/Record.java
acerdas84: I checked some steps and you are right, I saw libraries converting to a byte array again, but even in that case they do not return a byte array as a result (although you can probably get the bytes of the String they return). 
At first sight does not look trivial to do this, so I am taking your suggestion I will create a separate ticket for this.

src/main/java/eu/europeana/metis/sandbox/domain/Record.java
acerdas84: Added to [MET-2680](https://europeana.atlassian.net/browse/MET-2680) to be removed after the change

src/main/java/eu/europeana/metis/sandbox/service/workflow/OrderingService.java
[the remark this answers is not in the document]
acerdas84: [MET-2680](https://europeana.atlassian.net/browse/MET-2680)

src/main/java/eu/europeana/metis/sandbox/util/ZipFileReader.java
acerdas84: [MET-2680](https://europeana.atlassian.net/browse/MET-2680)

src/main/java/eu/europeana/metis/sandbox/service/workflow/MediaProcessingServiceImpl.java
```diff
@@ -1,13 +1,162 @@
...
+    try (MediaExtractor extractor = processorFactory.createMediaExtractor()) {
+      for (RdfResourceEntry entry : resourceEntries) {
+        var extraction = extractor.performMediaExtraction(entry);
+        if (extraction != null) {
+          results.add(extraction);
+        }
+      }
+    } catch (IOException | MediaProcessorException | MediaExtractionException e) {
+      throw new RecordProcessingException(record.getRecordId(), e);
+    }
```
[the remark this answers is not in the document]
acerdas84: Ohhhhh I understand, I have to rethink a couple of things then, I was not aware the same record could have more than one error related to it. Thanks for the feedback

src/main/java/eu/europeana/metis/sandbox/service/workflow/MediaProcessingServiceImpl.java
```diff
@@ -1,13 +1,152 @@
...
+      if (nonNull(extraction)) {
+        // Store thumbnails
+        storeThumbnails(record, extraction.getThumbnails(), recordErrors);
+        // Add result to RDF
+        rdfForEnrichment.enrichResource(extraction.getMetadata());
+      }
+    } catch (MediaExtractionException | IOException e) {
+      LOGGER.warn("Error while extracting media for record {}. ", record.getRecordId(), e);
+      // collect warn
+      recordErrors.add(new RecordError(e));
```
acerdas84, line 92: Now I am just collecting the errors and logging a warn when an entry has any issue

src/main/java/eu/europeana/metis/sandbox/service/workflow/MediaProcessingServiceImpl.java
```diff
@@ -1,13 +1,152 @@
...
+
+  private void storeThumbnails(Record record, List<Thumbnail> thumbnails,
+      List<RecordError> recordErrors) {
+    if (nonNull(thumbnails)) {
+      try {
+        thumbnailStoreService.store(thumbnails);
+      } catch (ThumbnailStoringException e) {
+        LOGGER.warn("Error while storing thumbnail for record {}. ", record.getRecordId(), e);
+        // collect warn
+        recordErrors.add(new RecordError(e));
```
acerdas84, line 104: Now I am just collecting the errors and logging a warn when an entry has any issue

src/main/java/eu/europeana/metis/sandbox/domain/Event.java
```diff
@@ -4,58 +4,50 @@
...
+import java.util.Collections;
+import java.util.List;
 
 /**
  * Event that contains a record and its processing details
  */
 public class Event {
 
   private final Record body;
+  private final List<RecordError> recordErrors;
```
acerdas84, line 16: Since same record can have several errors (like media processing) now we collect them in this list

src/main/java/eu/europeana/metis/sandbox/domain/RecordInfo.java
```diff
@@ -0,0 +1,32 @@
+package eu.europeana.metis.sandbox.domain;
+
+import static java.util.Objects.requireNonNull;
+
+import java.util.Collections;
+import java.util.List;
+
+public class RecordInfo {
```
acerdas84, line 8: This now is the return for each workflow step service, so that we can encapsulate non stoping errors that occured during processing

src/main/java/eu/europeana/metis/sandbox/service/record/RecordLogServiceImpl.java
```diff
@@ -23,21 +23,18 @@ public void logRecordEvent(Event recordEvent) {
...
     var record = recordEvent.getBody();
-    var eventError = recordEvent.getEventError();
-    String errorMessage = null;
-    String stackTrace = null;
-
-    if (eventError.isPresent()) {
-      errorMessage = eventError.get().getMessage();
-      stackTrace = eventError.get().getStackTrace();
-    }
+    // TODO address this in MET-2571
```
acerdas84, line 26: I have this on my radar for the other story I have in progress

src/main/java/eu/europeana/metis/sandbox/domain/Event.java
jochen-vermeulen: `Record` and `List<RecordError>` together are your new object `RecordInfo`. Should we maybe have a `RecordInfo` instance here instead? Or perhaps have this class be a subclass of `RecordInfo`?

acerdas84: Yes, I had it like that at the beggining then trying to follow [demeter law](https://medium.com/better-programming/demeters-law-don-t-talk-to-strangers-87bb4af11694) I change it (but it was a mistake to split the object here since all I needed to do was to add 2 get methods for each property) 
I will address this in [MET-2655](https://europeana.atlassian.net/browse/MET-2655)

PR #9, opened 2020-04-30
Feat/met 2571 errors report

acerdas84:
As a Sandbox user, I want to access errors reports for validation and media service

Also:

- Fix sonar issues
- Change datasetId from String to Integer

Example of report:

```json
{
  "progress": "TBD",
  "errors-report": [
    {
      "step": "external validation",
      "errors": [
        {
          "status": "error",
          "error-message": "cvc-complex-type.4: Attribute 'resource' must appear on element 'edm:object'.",
          "records": [
            "http://www.manuscriptorium.com/object/",
            "http://www.manuscriptorium.com/object/NLS2__RS_11_______45FULL2",
            "http://www.manuscriptorium.com/object/NLS2__RS_688______0U7I46D"
          ]
        },
        {
          "status": "error",
          "error-message": "cvc-complex-type.2.4.b: The content of element 'edm:ProvidedCHO' is not complete.",
          "records": [
            "/154/data_sounds_C0004X0034XX_0200",
            "/154/data_sounds_C1500X0418XX_0001"
          ]
        }
      ]
    },
    {
      "step": "media processing",
      "errors": [
        {
          "status": "warning",
          "error-message": "Problem while processing http://dom.lndb.lv/data/obj/file/244157",
          "records": [
            "57858"
          ]
        },
        {
          "status": "warning",
          "error-message": "Problem while processing http://dom.lndb.lv/data/obj/58356.html",
          "records": [
            "58356"
          ]
        },
        {
          "status": "warning",
          "error-message": "Problem while processing http://dom.lndb.lv/data/obj/57858.html",
          "records": [
            "57858"
          ]
        },
        {
          "status": "warning",
          "error-message": "Problem while processing http://dom.lndb.lv/data/obj/file/252571",
          "records": [
            "58339"
          ]
        },
        {
          "status": "warning",
          "error-message": "Problem while processing http://dom.lndb.lv/data/obj/58211.html",
          "records": [
            "58211"
          ]
        },
        {
          "status": "warning",
          "error-message": "Problem while processing http://dom.lndb.lv/data/obj/58339.html",
          "records": [
            "58339"
          ]
        },
        {
          "status": "warning",
          "error-message": "Problem while processing http://dom.lndb.lv/data/obj/file/244195",
          "records": [
            "58211"
          ]
        },
        {
          "status": "warning",
          "error-message": "Problem while processing http://dom.lndb.lv/data/obj/file/252270",
          "records": [
            "58356"
          ]
        },
        {
          "status": "warning",
          "error-message": "Problem while processing http://dom.lndb.lv/data/obj/58250.html",
          "records": [
            "58250"
          ]
        },
        {
          "status": "warning",
          "error-message": "Problem while processing http://dom.lndb.lv/data/obj/file/250860",
          "records": [
            "58250"
          ]
        }
      ]
    }
  ]
}
```


src/main/java/eu/europeana/metis/sandbox/config/amqp/AmqpConfiguration.java
```diff
@@ -125,97 +125,28 @@ Declarables deadQueues() {
...
-        .setEnriched(enrichedDlq)
-        .setMediaProcessed(mediaProcessedDlq)
-        .setIndexed(indexedDlq);
-    return getDeclarables(dlqs);
+    return getDeclarables(createdDlq, externalValidatedDlq, transformedDlq,
+        normalizedDlq, internalValidatedDlq, enrichedDlq, mediaProcessedDlq, indexedDlq);
   }
 
-  private Declarables getDeclarables(Queues qs) {
+  private Declarables getDeclarables(String created, //NOSONAR
```
acerdas84, line 138: Sonar complains about too many parameters, but if I split this into 2 methods and use the instance variables, the it will complain about duplicate code. So from those I pick this one and ignore Sonar

src/main/java/eu/europeana/metis/sandbox/consumer/workflow/CreatedConsumer.java
```diff
@@ -48,7 +48,7 @@ public void validateExternal(Event input) {
       var status = record.getErrors().isEmpty() ? Status.SUCCESS : Status.WARN;
       output = new Event(record, Step.VALIDATE_EXTERNAL, status);
     } catch (RecordProcessingException ex) {
-      LOGGER.error(ex.getMessage(), ex);
+      LOGGER.error("Exception while performing external validation step. ", ex);
```
acerdas84, line 51: As Sonar remind me the message from the exception is already logged by the LOGGER

src/main/java/eu/europeana/metis/sandbox/controller/DatasetController.java
```diff
@@ -37,19 +37,23 @@
...
   }
 
   @ApiOperation("Process the given dataset")
   @PostMapping(value = "dataset/{name}/process", produces = APPLICATION_JSON_VALUE)
   @ResponseStatus(HttpStatus.ACCEPTED)
   public DatasetIdDto processDataset(
       @ApiParam(value = "name of the dataset", required = true) @PathVariable(value = "name") String datasetName,
-      @ApiParam(value = "country of the dataset", required = true, defaultValue = "NETHERLANDS") @RequestParam Country country,
-      @ApiParam(value = "language of the dataset", required = true, defaultValue = "NL") @RequestParam Language language,
+      @ApiParam(value = "country of the dataset", required = true, defaultValue = "Netherlands") @RequestParam Country country,
```
acerdas84, line 55: Now the list of countries and languages will come from the variable value of the enum

src/main/java/eu/europeana/metis/sandbox/domain/Record.java
```diff
@@ -22,21 +21,21 @@
...
-  private Record(String recordId, String datasetId, String datasetName,
+  private Record(String recordId, Integer datasetId, String datasetName,
       Country country, Language language, byte[] content) {
     this.recordId = recordId;
     this.datasetId = datasetId;
     this.datasetName = datasetName;
     this.country = country;
     this.language = language;
-    this.content = content;
+    this.content = content; //NOSONAR
```
acerdas84, line 38: Sonar complains because 

> Mutable members should not be stored or returned directly

But I am creating a copy of the content in the builder, so here we will have a copy of content

src/main/java/eu/europeana/metis/sandbox/domain/Record.java
```diff
@@ -101,25 +100,25 @@ public Language getLanguage() {
...
-   * <br />
-   * We are not making a copy of it because it
-   * is expensive and Record object is expected to be use as a non mutable object
+   * @implNote Overwriting this field contents after construction could cause problems. <br /> We
+   * are not making a copy of it because it is expensive and Record object is expected to be use as
+   * a non mutable object
    */
   public byte[] getContent() {
-    return this.content;
+    return this.content; //NOSONAR
```
acerdas84, line 108: Sonar complains because

> Mutable members should not be stored or returned directly

I am aware of this, but returning a copy of the byte array will be expensive.
But I started thinking if we include methods in metis libraries to handle byte array directly then same bug is going to start popping everywhere, I am starting to guess the reason in metis libraries to treat the record as a String is because of the immutability of it which makes it easier to work in trade for transform to and from byte array (which most metis libraries do). 
I started switching to byte array mostly because when I publish/consume messages from rabbit, the record xml representation travels as a byte[] but I treat it in the application as a String, this is actually common, because when you send content through Rabbit, Spring AMQP will set it as a byte[] and if you do not specify a custom MessageConverter then the transformation from and to a String will happen under the hoods, same happens with Kafka. 
So I am thinking between going back to Strings to take advantage of its immutability and pay the price of transformations between publish/consume in Rabbit (which is common anyways) and keep using metis libraries as they are now (which most of them take String as the record parameter) because transform back and forth String and byte[] is a concern Spring AMQP. 
Or we can push to change it all to byte[] but we need to understand that the content of a record could be altered and also I am not sure if metis libraries could have an impact on that change.
Anyways there is a story to address the change, I just started thinking if it is worth to do it, and what impact it could have.

src/main/java/eu/europeana/metis/sandbox/common/exception/RecordValidationException.java
```diff
@@ -14,7 +14,7 @@
 
   public RecordValidationException(String message, String recordId, String nodeId) {
     super(format("Record: %s Node: %s Message: %s ",
-        recordId, nodeId, message), recordId, null);
+        recordId, nodeId, message), recordId, message, null);
     this.nodeId = nodeId;
```
jochen-vermeulen, line 18: Do you anticipate a use for this class variable? I don't think the getter is used anywhere.

src/main/java/eu/europeana/metis/sandbox/config/amqp/AmqpConfiguration.java
jochen-vermeulen: Ok, but I would say to add a specific sonar issue ID here, so that if another issue is triggered, we don't miss it. You can do this like so (replace the string argument by the ID of your specific sonarQube issue):
```
@SuppressWarnings("findbugs:RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE")
```
And perhaps put some comments in the code to explain the reason.

src/main/java/eu/europeana/metis/sandbox/domain/Record.java
jochen-vermeulen: Same as above: use the suppressed warnings annotation.

src/main/java/eu/europeana/metis/sandbox/domain/Record.java
jochen-vermeulen: I understand. But perhaps we could choose a middle path. Instead of byte arrays, we could pass InputStreams to the library as well. If we change this method to return a bytearrayinputstream we get rid of the warning, and we can change the libraries to accept this instead of (or in addition to) byte arrays.
What do you think?
In the end, I think in this case performance considerations should trump code quality issues.

src/main/java/eu/europeana/metis/sandbox/common/amqp/RecordMessageConverter.java
```diff
@@ -72,7 +72,7 @@ public Object fromMessage(Message message) {
     byte[] content = message.getBody();
     MessageProperties properties = message.getMessageProperties();
     String recordId = properties.getHeader(RECORD_ID);
-    String datasetId = properties.getHeader(DATASET_ID);
+    Integer datasetId = properties.getHeader(DATASET_ID);
```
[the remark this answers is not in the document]
stzanakis: Good find! Well the best would be indeed Integer. But to keep the consistency with metis, because of the legacy ids, we should keep it as String I would say..

src/main/java/eu/europeana/metis/sandbox/config/amqp/AmqpConfiguration.java
acerdas84: Nice, I will change it

src/main/java/eu/europeana/metis/sandbox/domain/Record.java
acerdas84: I think `ByteArrayInputStream` sounds good, then metis libraries could take it instead of String or byte[]. And thinking again we could reconsider returning a copy of the byte[] (when byte[] is needed), at the end it will still be better than doing `String.getBytes()` 

Also I think while it will be faster we are actually not sure of the performance gains, we do not have measures of what we are winning with these changes (or at least I am not sure, maybe you already notice this in the past with ecloud team and have info on the improvements). So I will add to [MET-2680](https://europeana.atlassian.net/browse/MET-2680) a comment on running tests to measure the performance wins with these changes.
At the end I think we consider this because we will win something, but I want to understand what is that something we are winning

src/main/java/eu/europeana/metis/sandbox/common/amqp/RecordMessageConverter.java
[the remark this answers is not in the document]
acerdas84: Yes, since I am generating the value from the database I set it as an Integer, because it will not be anything else, but what are we winning with keeping it as a String? It might be confusing to think it could be a String, and since this is a new application why not going with the right type?

src/main/java/eu/europeana/metis/sandbox/common/exception/RecordValidationException.java
acerdas84: Not really, I have not found a use for it, but when defining custom exceptions I try to provide relevant info from them, in this case the node id. On the other hand I have no problem removing it, I just did not want my use of it to drive the way it should be defined (if that makes sense)

src/main/java/eu/europeana/metis/sandbox/common/amqp/RecordMessageConverter.java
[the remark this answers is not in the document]
acerdas84: Reverted to String

PR #10, opened 2020-05-12
Feat/met 2546 index and preview

acerdas84:
Be able to index and preview data after media service


pom.xml
```diff
@@ -145,6 +145,34 @@
...
+			<exclusions>
+				<exclusion>
+					<groupId>org.apache.logging.log4j</groupId>
+					<artifactId>log4j-slf4j-impl</artifactId>
+				</exclusion>
+			</exclusions>
+		</dependency>
+
+		<dependency>
+			<groupId>eu.europeana.corelib</groupId>
```
acerdas84, line 161: Took me time to figure but this one is required by indexing, but the enrich pom explicitly excludes it, so I had to add it again

PR #12, opened 2020-05-14
Debt/met 2680 use byte[] and fix tests

acerdas84:
* Apply changes to sandbox workflow services in package eu.europeana.metis.sandbox.service.workflow to use byte array methods instead of String ones
* Remove method getContentString() from Record class and add needed changes


pom.xml
```diff
@@ -78,12 +78,6 @@
...
 			<version>${version.metis}</version>
-			<exclusions>
-				<exclusion>
-					<groupId>org.apache.logging.log4j</groupId>
-					<artifactId>log4j-slf4j-impl</artifactId>
-				</exclusion>
-			</exclusions>
 		</dependency>
 
 		<dependency>
```
jochen-vermeulen, line 83: How about this dependency? Should the exclusion not be removed here either?

src/main/java/eu/europeana/metis/sandbox/service/workflow/EnrichmentServiceImpl.java
```diff
@@ -24,10 +23,10 @@ public EnrichmentServiceImpl(EnrichmentWorker enrichmentWorker) {
   public RecordInfo enrich(Record record) {
     requireNonNull(record, "Record must not be null");
 
-    String result;
+    byte[] result;
     try {
-      result = enrichmentWorker.process(record.getContentString());
-    } catch (DereferenceOrEnrichException | JiBXException | UnsupportedEncodingException e) {
+      result = enrichmentWorker.process(record.getContentInputStream());
+    } catch (DereferenceOrEnrichException | JiBXException e) {
```
jochen-vermeulen, line 29: JiBXException is never thrown in the try block.

pom.xml
acerdas84: That is what I thought, and I did try, but that service depends on `corelib-edm-definitions`, which [parent](https://github.com/europeana/corelib/blob/develop/pom.xml) has the dependency

jochen-vermeulen: I see, thanks. It should therefore be excluded there, not here. I'll change it in metis.

src/main/java/eu/europeana/metis/sandbox/service/workflow/EnrichmentServiceImpl.java
acerdas84: That is in fact correct, I will remove it

PR #13, opened 2020-05-18
MET-2727: Add portalUrl when workflow execution is done

acerdas84:
Looks like this

```json
{
  "portalUrl": "https://metis-sandbox-preview-api-test-portal.eanadev.org/portal/search?view=grid&q=edm_datasetName:12_test*",
  "status": "completed",
  "total-records": 232,
  "processed-records": 232,
  "progress-by-step": [
    {
      "step": "import",
      "total": 232,
      "success": 232,
...
```

### DIFF
--- a/src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetReportServiceImpl.java
+++ b/src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetReportServiceImpl.java
@@ -19,12 +19,15 @@ import eu.europeana.metis.sandbox.repository.DatasetRepository;
 import eu.europeana.metis.sandbox.repository.RecordErrorLogRepository;
 import eu.europeana.metis.sandbox.repository.RecordLogRepository;
 import eu.europeana.metis.sandbox.repository.projection.ErrorLogView;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.util.Comparator;
 import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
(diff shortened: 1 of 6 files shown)


src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetReportServiceImpl.java
```diff
@@ -85,7 +95,18 @@ public DatasetInfoDto getReport(String datasetId) {
...
+        stepsInfo);
+  }
+
+  private String getPortalUrl(DatasetEntity dataset, long completedRecords) {
+    long recordsQty = dataset.getRecordsQuantity();
+    if (recordsQty != completedRecords) {
+      return PORTAL_MSG;
+    }
+    var datasetId = dataset.getDatasetId() + SEPARATOR + dataset.getDatasetName() + SUFFIX;
+    return portalUrl + URLEncoder.encode(datasetId, StandardCharsets.UTF_8);
```
acerdas84, line 109: This generates something like
`http://preview-metis/portal/search?view=grid&q=edm_datasetName:12_test*`
But the portal by default filter out content tier 0 results, should we be including them? I know what to add to the url but just want to confirm if it makes sense

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetReportServiceImpl.java
```diff
@@ -19,19 +19,28 @@
...
 import java.util.Optional;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
 @Service
 class DatasetReportServiceImpl implements DatasetReportService {
 
   private static final int FIRST = 0;
+  private static final String PORTAL_MSG = "Generates after process is completed";
```
jochen-vermeulen, line 38: 1. Maybe make this more readable? "A review URL will be generated when the dataset has finished processing"?
2. Should we also include a message for if there are no successful records?

PR #15, opened 2020-05-19
Feat/met 2733 publish records

acerdas84:
Adding one more step for publishing the dataset records

### DIFF
--- a/src/main/java/eu/europeana/metis/sandbox/common/IndexEnv.java
+++ b/src/main/java/eu/europeana/metis/sandbox/common/IndexEnv.java
@@ -1,6 +1,6 @@
 package eu.europeana.metis.sandbox.common;
 
-public enum Index {
+public enum IndexEnv {
   PREVIEW,
   PUBLISH
 }
--- a/src/main/java/eu/europeana/metis/sandbox/config/IndexingConfig.java
+++ b/src/main/java/eu/europeana/metis/sandbox/config/IndexingConfig.java
@@ -20,48 +20,112 @@ import org.springframework.context.annotation.Configuration;
 @Configuration
 class IndexingConfig {
 
-  @Value("${sandbox.indexing.mongo.hosts}")
-  private String[] mongoHosts;
+  @Value("${sandbox.preview.mongo.hosts}")
+  private String[] mongoPreviewHosts;
 
-  @Value("${sandbox.indexing.mongo.ports}")
-  private int[] mongoPorts;
+  @Value("${sandbox.preview.mongo.ports}")
+  private int[] mongoPreviewPorts;
 
-  @Value("${sandbox.indexing.mongo.authentication-db:#{null}}")
-  private String mongoAuthenticationDb;
+  @Value("${sandbox.preview.mongo.authentication-db:#{null}}")
+  private String mongoPreviewAuthenticationDb;
 
-  @Value("${sandbox.indexing.mongo.username:#{null}}")
-  private String mongoUsername;
+  @Value("${sandbox.preview.mongo.username:#{null}}")
+  private String mongoPreviewUsername;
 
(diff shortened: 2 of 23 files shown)


src/main/java/eu/europeana/metis/sandbox/config/IndexingConfig.java
```diff
@@ -20,62 +20,124 @@
...
     checkArgument(isNotEmpty(mongoHosts), "Mongo hosts must be provided ");
     checkArgument(isNotEmpty(mongoPorts), "Mongo ports must be provided ");
     checkArgument(isNotEmpty(solrHosts), "Solr hosts must be provided ");
 
     var settings = new IndexingSettings();
 
-    if (isNull(mongoEnableSSL)) {
-      mongoEnableSSL = Boolean.FALSE;
-    }
+    boolean enableSSL = !isNull(mongoEnableSSL) && mongoEnableSSL;
```
stzanakis, line 136: How about:
boolean enableSSL = Boolean.TRUE.equals(mongoEnableSSL);
?
Or even better use Boolean.TRUE.equals(mongoEnableSSL); on the setAllProperties parameter directly below, instead of creating a new variable.

src/main/java/eu/europeana/metis/sandbox/common/IndexEnv.java
```diff
@@ -0,0 +1,6 @@
+package eu.europeana.metis.sandbox.common;
+
+public enum IndexEnv {
```
stzanakis, line 3: Can we have full wording here?
"IndexEnvironment"

src/main/java/eu/europeana/metis/sandbox/config/IndexingConfig.java
```diff
@@ -20,62 +20,124 @@
...
+    return getIndexer(mongoPublishHosts, mongoPublishPorts, mongoPublishDb,
+        mongoPublishAuthenticationDb,
+        mongoPublishUsername, mongoPublishPassword, mongoPublishEnableSSL, solrPublishHosts,
+        zookeeperPublishHosts, zookeeperPublishPorts, zookeeperPublishChroot,
+        zookeeperPublishDefaultCollection, zookeeperPublishTimeoutInSecs);
+  }
+
+  //Suppress: Methods should not have too many parameters warning
+  //We are okay with this method to ease configuration
+  @SuppressWarnings("squid:S107")
```
stzanakis, line 123: Not critical, but we could perhaps bundle the mongo config and solr/zookeeper config?

acerdas84: I am not sure to understand the suggestion. Could you expand? maybe an example

stzanakis: Well I was thinking something in the line like:
getIndexer(getMongoConfiguration(mongoPublishHosts, mongoPublishPorts, mongoPublishDb,
        mongoPublishAuthenticationDb, mongoPublishUsername, mongoPublishPassword), getSolrConfiguration(solrPublishHosts,
        zookeeperPublishHosts, zookeeperPublishPorts, zookeeperPublishChroot,
        zookeeperPublishDefaultCollection, zookeeperPublishTimeoutInSecs))

And those two methods would return two classes that encapsulate the corresponding properties..

Was just an idea of how to fix that warning, but we don't have to follow that, as you prefer.

acerdas84: I can do it, I will apply it in the ticket I am working now if that is ok
Thanks

stzanakis: Yeah that's fine!

PR #16, opened 2020-05-26
Feat/met 2554 purge dataset

acerdas84:
Purge datasets that are more than a week old


src/main/java/eu/europeana/metis/sandbox/entity/ThumbnailEntity.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+@Entity
+@Table(name = "thumbnail")
+public class ThumbnailEntity {
+
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  private Long id;
+
+  private String datasetId;
```
jochen-vermeulen, line 20: Should this not be a many to one mapping to dataset? Or are we not in need of relationships here? Could help maintain the integrity?

src/main/java/eu/europeana/metis/sandbox/repository/DatasetRepository.java
```diff
@@ -1,8 +1,19 @@
...
 public interface DatasetRepository extends JpaRepository<DatasetEntity, Integer> {
 
+  /**
+   * Get a list of datasets created before specified date
+   *
+   * @param date must not be null
+   * @return list of dataset ids
+   * @see DatasetIdView
+   */
+  List<DatasetIdView> getByCreatedDateBefore(LocalDateTime date);
```
jochen-vermeulen, line 18: Just curiosity: where do you define the query that does this operation?

acerdas84: I don't, [Spring Data](https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#jpa.query-methods.query-creation) defines it

src/main/java/eu/europeana/metis/sandbox/entity/ThumbnailEntity.java
acerdas84: No I am not using them here, I thought of this in kind of microservice scaling. So for example in the future you could put all the media processing step and thumbnails logic in another service, with its own database, and it will keep track of their own domain info, anyways I could add the relationship in the database and that will help with integrity, but rather leaving it out of the entities themselves

src/main/java/eu/europeana/metis/sandbox/repository/DatasetRepository.java
jochen-vermeulen: I see. Both cool and scary... :)
Could you perhaps put that link in the comments there, so that we can find this?

src/main/java/eu/europeana/metis/sandbox/entity/ThumbnailEntity.java
jochen-vermeulen: Alright. Sounds reasonable.

src/main/java/eu/europeana/metis/sandbox/repository/DatasetRepository.java
acerdas84: Yes I will. 
Also you might notice that for other methods in other repositories with `deleteByDatasetId` I defined the query with the @Query annotation, the reason for that is that if you let Spring Data to define that one for you it will first perform a select of the records based on that property, and then delete one by one (because when doing a delete a list of deleted records is returned), and that takes more time than just what I wanted which was `delete from table where datasetId =` but in trade only the number of deleted records can be returned, which I do not use anyways so you see my returns are void.
So in my experience when using this feature, test locally changing `logback-spring.xml` with this 
``` <logger name="org.hibernate" level="DEBUG"/>```
So you can verify how the generated query will look like. And if you find out is not optimal then you can define your own using the @Query annotation

PR #17, opened 2020-05-26
Feat/met 2740 test sandbox

acerdas84:
Filtering duplicate record ids

Also:

- Added flags to enable or disable queue listeners in application.yml (default value is enabled)

### DIFF
--- a/src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetGeneratorServiceImpl.java
+++ b/src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetGeneratorServiceImpl.java
@@ -2,6 +2,7 @@ package eu.europeana.metis.sandbox.service.dataset;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.toSet;
 
 import eu.europeana.metis.sandbox.common.locale.Country;
 import eu.europeana.metis.sandbox.common.locale.Language;
@@ -10,7 +11,6 @@ import eu.europeana.metis.sandbox.domain.Record;
 import eu.europeana.metis.sandbox.service.util.XmlRecordProcessorService;
 import java.io.ByteArrayInputStream;
 import java.util.List;
-import java.util.stream.Collectors;
 import org.springframework.stereotype.Service;
 
 @Service
@@ -32,6 +32,7 @@ class DatasetGeneratorServiceImpl implements DatasetGeneratorService {
     requireNonNull(records, "Records must not be null");
     checkArgument(!records.isEmpty(), "Records must not be empty");
 
+    var totalRecords = records.size();
     var recordObjectList = records.stream()
         .map(ByteArrayInputStream::readAllBytes)
         .map(record -> Record.builder()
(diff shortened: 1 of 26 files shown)


src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetGeneratorServiceImpl.java
```diff
@@ -42,8 +43,8 @@ public Dataset generate(String id, String name, Country country, Language langua
             .language(language)
             .content(record)
             .build())
-        .collect(Collectors.toList());
+        .collect(toSet());
 
-    return new Dataset(id, recordObjectList);
+    return new Dataset(id, recordObjectList, totalRecords - recordObjectList.size());
```
jochen-vermeulen, line 48: I don't think this is the right way to count duplicates. I think we should be looking at the record IDs alone (and the equals method in Record look at other properties as well). Perhaps two records with the same ID but different languages exist, but they should still be counted as duplicates I think.

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetServiceImpl.java
```diff
@@ -38,15 +41,27 @@ public String createDataset(String datasetName, Country country, Language langua
...
+    // if there are duplicate records in the original list
+    if (dataset.getRecords().size() < records.size()) {
+      // adjust records qty to be equal to non duplicate records
+      entity.setRecordsQuantity(dataset.getRecords().size());
+      try {
+        datasetRepository.save(entity);
+      } catch (Exception e) {
+        throw new ServiceException("Error updating the dataset: " + e.getMessage(), e);
+      }
+    }
```
jochen-vermeulen, line 62: Unless I'm mistaken, I think you can do this check before saving the entity the first time. I think that would be better... first create the entity completely with all counters and then save it once?

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetGeneratorServiceImpl.java
acerdas84: Yes I will update the equals method for the Record class

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetServiceImpl.java
acerdas84: In line 50 I need the datasetId to call generatorService, which is in charge of creating the Set<Records> (by reading the List< ByteArrayInputStream>). Each Record object contains the datasetId, then after calling generatorService I know how many records are duplicates, that is why I update the entity here, but I was not able to produce all counters at first.

jochen-vermeulen: You're right, I missed that. But still there must be a way... but perhaps that's not the most important thing to work on right now. This only occurs when there are duplicates, which should not happen too often.

acerdas84: Yes I thought it should not be often,  but I will create a story to not forget about this

acerdas84: [MET-2760](https://europeana.atlassian.net/browse/MET-2760)

PR #18, opened 2020-05-28
Feat/met 2740 fix null error message

acerdas84:
Changes coming from testing

### DIFF
--- a/src/main/java/eu/europeana/metis/sandbox/config/amqp/AmqpConfiguration.java
+++ b/src/main/java/eu/europeana/metis/sandbox/config/amqp/AmqpConfiguration.java
@@ -140,14 +140,14 @@ class AmqpConfiguration {
 
   @Bean
   Declarables bindings() {
-    return getDeclarables(createdQueue, externalValidatedQueue, transformedQueue,
+    return getDeclarables(exchange, createdQueue, externalValidatedQueue, transformedQueue,
         normalizedQueue, internalValidatedQueue, enrichedQueue, mediaProcessedQueue,
         previewedQueue, publishedQueue);
   }
 
   @Bean
   Declarables dlqBindings() {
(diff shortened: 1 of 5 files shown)


src/main/java/eu/europeana/metis/sandbox/config/amqp/AmqpConfiguration.java
```diff
@@ -140,22 +140,22 @@ Declarables deadQueues() {
 
   @Bean
   Declarables bindings() {
-    return getDeclarables(createdQueue, externalValidatedQueue, transformedQueue,
+    return getDeclarables(exchange, createdQueue, externalValidatedQueue, transformedQueue,
```
acerdas84, line 143: Big bug I had that it was preventing fail messages going to dead letter queue

src/main/java/eu/europeana/metis/sandbox/domain/RecordError.java
```diff
@@ -21,14 +22,15 @@
    */
   public RecordError(RecordProcessingException exception) {
     requireNonNull(exception, "Exception must not be null");
-    this.message = exception.getReportMessage();
+    this.message = Optional.ofNullable(exception.getReportMessage())
+        .orElse("No message. Report to service provider");
```
acerdas84, line 26: Avoiding error message to be null, after dealing with some tests I realize it was a mistake not to check for this

src/main/resources/logback-spring.xml
```diff
@@ -15,7 +15,7 @@
 
   <logger name="eu.europeana" level="WARN"/>
   <logger name="eu.europeana.metis.sandbox" level="INFO"/>
-  <logger name="org.springframework" level="ERROR"/>
+  <logger name="org.springframework" level="WARN"/>
```
acerdas84, line 18: After doing troubleshooting of messages going to dead letter queue, I realize this should be at least in WARN level

PR #23, opened 2021-02-22
MET-2830 Update sandbox to match changes with eCloud

JoanaCMS:


src/main/java/eu/europeana/metis/sandbox/service/workflow/MediaProcessingServiceImpl.java
```diff
@@ -76,17 +78,51 @@ public RecordInfo processMedia(Record record) {
...
+      RdfResourceEntry mainThumbnail = rdfDeserializer
+          .getMainThumbnailResourceForMediaExtraction(inputRdf);
+      extractedThumbnail = extractor.performMediaExtraction(mainThumbnail, false);
+    } catch (RdfDeserializationException | MediaExtractionException e) {
+      LOGGER.warn("Error while extracting media for record {}. ", record.getRecordId(), e);
+      recordErrors.add(new RecordError(new RecordProcessingException(record.getRecordId(), e)));
+    }
+
+    boolean hasThumbnail =
+        extractedThumbnail != null && !extractedThumbnail.getThumbnails().isEmpty();
```
jochen-vermeulen, line 101: Here you should also call `rdfForEnrichment.enrichResource` (if `extractedThumbnail` is not null). In fact, the result is not just a thumbnail, but actually a media extraction that may or may not have a thumbnail (but if it does it is the main thumbnail).

src/main/java/eu/europeana/metis/sandbox/service/workflow/MediaProcessingServiceImpl.java
```diff
@@ -55,17 +54,20 @@ public RecordInfo processMedia(Record record) {
...
-      for (RdfResourceEntry entry : resourceEntries) {
-        processResourceEntry(record, rdfForEnrichment, extractor, entry, recordErrors);
-      }
+      // Get main thumbnail
+      boolean hasThumbnail = processMainThumbnailResource(rdfDeserializer, inputRdf, extractor,
+          record, recordErrors);
+
+      // Get resource entries
+      processResourceEntry(record, rdfForEnrichment, extractor, inputRdf, hasThumbnail,
+          rdfDeserializer, recordErrors);
```
jochen-vermeulen, line 67: Once you have solved the other two comments, you should find that the two methods you created are quite similar. Perhaps then you can merge them.

jochen-vermeulen (changes requested):

src/main/java/eu/europeana/metis/sandbox/service/workflow/MediaProcessingServiceImpl.java
JoanaCMS: Indeed, the methods became very similar. Created a new method that should work for both main thumbnail and remaining resources

src/main/java/eu/europeana/metis/sandbox/service/workflow/MediaProcessingServiceImpl.java
JoanaCMS: Done

src/main/java/eu/europeana/metis/sandbox/service/workflow/MediaProcessingServiceImpl.java
```diff
@@ -76,17 +78,51 @@ public RecordInfo processMedia(Record record) {
...
-        // Store thumbnails
-        storeThumbnails(record, extraction.getThumbnails(), recordErrors);
+      MediaExtractor extractor, byte[] inputRdf, boolean hasThumbnail,
+      RdfDeserializer rdfDeserializer, List<RecordError> recordErrors) {
+
+    try {
+      List<RdfResourceEntry> remainingResources = rdfDeserializer
+          .getRemainingResourcesForMediaExtraction(inputRdf);
+      for (RdfResourceEntry entry : remainingResources) {
+        ResourceExtractionResult extraction = extractor.performMediaExtraction(entry, hasThumbnail);
```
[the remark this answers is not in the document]
JoanaCMS: Done

src/main/java/eu/europeana/metis/sandbox/service/workflow/MediaProcessingServiceImpl.java
```diff
@@ -76,21 +82,45 @@ public RecordInfo processMedia(Record record) {
...
+
+    try{
+
+      // Check if the process for main thumbnail has happened.
+      // If not, then get main thumbnail
+      if(!hasRanMainThumbnail) {
+        extraction = extractor.performMediaExtraction(resourceToProcess, false);
+      } else {
+        extraction = extractor.performMediaExtraction(resourceToProcess, gotMainThumbnail);
+      }
```
jochen-vermeulen, line 105: You could get rid of both this if statement and the hasRanMainThumbnail parameter for this method by replacing this with just line 104. Did you have a reason for not going that way?

src/main/java/eu/europeana/metis/sandbox/service/workflow/MediaProcessingServiceImpl.java
```diff
@@ -76,21 +82,45 @@ public RecordInfo processMedia(Record record) {
...
+
+      // Check if extraction for media was successful
+      successful = extraction != null && !extraction.getThumbnails().isEmpty();;
```
jochen-vermeulen, line 108: Be careful here: success is not defined by whether or not there are thumbnails! Some resources never have thumbnails (such as audio files). You will need an extra if statement within the if below to check for the availability of thumbnails.

src/main/java/eu/europeana/metis/sandbox/service/workflow/MediaProcessingServiceImpl.java
```diff
@@ -1,5 +1,6 @@
 package eu.europeana.metis.sandbox.service.workflow;
 
+import static java.util.Objects.checkFromIndexSize;
```
jochen-vermeulen, line 3: Unused import. :)

src/main/java/eu/europeana/metis/sandbox/service/workflow/MediaProcessingServiceImpl.java
JoanaCMS: Yes, I did had a reason when I first started refactoring the code. I was trying to avoid repeating the same step when there was no main thumbnail. 
You are right, it is not needed this `if` statement anymore. I will fix this

src/main/java/eu/europeana/metis/sandbox/service/workflow/MediaProcessingServiceImpl.java
JoanaCMS: Good point, I will fix this

src/main/java/eu/europeana/metis/sandbox/service/workflow/MediaProcessingServiceImpl.java
JoanaCMS: Oops. Removed!

PR #24, opened 2021-03-08
MET-3429 Report not printing all steps fix

JoanaCMS:


acerdas84:
You might want to consider at some point to upgrade the spring boot version, it should be straightforward 🙂

src/main/java/eu/europeana/metis/sandbox/config/SandboxConfig.java
```diff
@@ -55,10 +55,13 @@
   @Value("${sandbox.enrichment.enrichment-url}")
   private String enrichmentServiceUrl;
 
-  private String defaultXsltUrl;
-
+  //TODO: 04-03-2021 We should remove this configuration once
+  //TODO: XsltTransformation in metis is fixed
+  @Value("${sandbox.url.edm-sorted-url}")
   private String edmSorterUrl;
```
jochen-vermeulen, line 61: I think now the method `edmSorterUrl()` is no longer used.

src/main/java/eu/europeana/metis/sandbox/config/SandboxConfig.java
```diff
@@ -74,7 +77,7 @@ XPathFactory xPathFactory() {
   @Bean
   @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
   XsltTransformer xsltEdmSorter() throws IOException, TransformationException {
```
jochen-vermeulen, line 79: According to my IDE this method no longer needs to throw `IOException`.

src/main/java/eu/europeana/metis/sandbox/config/SandboxConfig.java
jochen-vermeulen: Also, the file referenced in that method (`classpath:edm/edm.xsd.sorter.xsl`) can be removed, I think?

src/test/resources/application.yml
```diff
@@ -25,6 +25,8 @@ spring:
 sandbox:
   cors:
     mapping: '*'
+  url:
+    edm-sorted-url: http://ftp.eanadev.org/schema_zips/edm_sorter.xsl
```
jochen-vermeulen, line 29: Is it wise to have this dependency in our test suite? Perhaps it's better to have this file locally (or absent and mock the code that uses it).

jochen-vermeulen:
> You might want to consider at some point to upgrade the spring boot version, it should be straightforward slightly_smiling_face

@acerdas84 Thanks! We created a ticket for this.

PR #25, opened 2021-03-12
MET-3419 Use new configuration from metis library

JoanaCMS:


src/main/java/eu/europeana/metis/sandbox/config/IndexingConfig.java
```diff
@@ -102,6 +102,9 @@
   @Value("${sandbox.publish.solr.zookeeper.timeout:#{null}}")
   private Integer zookeeperPublishTimeoutInSecs;
 
+  @Value("${sandbox.application-name}")
```
stzanakis, line 105: This is not at the right place, it should be part of the mongo configuration and in fact there should be one name for the preview indexer and one for the publish indexer.

JoanaCMS: I see what you mean. I have fixed it

src/main/java/eu/europeana/metis/sandbox/config/IndexingConfig.java
```diff
@@ -45,6 +45,9 @@
   @Value("${sandbox.preview.mongo.db}")
   private String mongoPreviewDb;
 
+  @Value("${sandbox.preview.mongo.application-name}")
```
stzanakis, line 48: Okay it looks good now, but I think we need to have it get default value of null like the other fields otherwise Spring might interpret it as empty String, did you check that?
(Same for the publish one)

acerdas84: When spring does not find a match for the property it will throw an exception with a message similar to:
```java
nested exception is java.lang.IllegalArgumentException: Could not resolve placeholder 'test.myprop.here' in value "${test.myprop.here}"
```
Unless there is a property with that name but with an empty value, in that case spring will assign an empty string value

JoanaCMS: Thank you! That is correct, I have just tested as well. I will add the default null values.

PR #26, opened 2021-03-17
MET-3489 Create new GET requests

JoanaCMS:


src/main/java/eu/europeana/metis/sandbox/controller/DatasetController.java
```diff
@@ -54,6 +59,7 @@
   private final ZipService zipService;
   private final DatasetService datasetService;
   private final DatasetReportService reportService;
+  private DatasetController test;
```
stzanakis, line 62: I think this is some leftovers?

src/main/java/eu/europeana/metis/sandbox/controller/DatasetController.java
```diff
@@ -93,4 +99,66 @@ public DatasetInfoDto retrieveDataset(
...
+  })
+  @GetMapping(value = "dataset/languages", produces = APPLICATION_JSON_VALUE)
+  @ResponseStatus(HttpStatus.OK)
+  @ResponseBody
+  public List<LanguageModel> getDatasetsLanguages(){
+    return Language.getLanguageListSortedByName().stream().map(LanguageModel::new)
+        .collect(Collectors.toList());
+  }
+
+  private static class CountryModel {
```
stzanakis, line 139: I can see this is basically a copy from metis-framework.
We've decided for any object that is for viewing to suffix it with "View" instead of "Model" and use "Model" for db objects.
I know it's metis-framework this is not updated yet, but perhaps here we can update already

src/main/java/eu/europeana/metis/sandbox/controller/DatasetController.java
JoanaCMS: You are correct, it is some leftovers, will remove it

src/main/java/eu/europeana/metis/sandbox/controller/DatasetController.java
JoanaCMS: So if I understood correctly, instead of `CountryModel` you say to change it to `CountryView`, correct?
If that is it, then I will updated it

PR #27, opened 2021-04-08
MET-3536 Get name language country and creation date

JoanaCMS:


src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetReportServiceImpl.java
```diff
@@ -11,7 +11,10 @@
 import eu.europeana.metis.sandbox.common.Step;
 import eu.europeana.metis.sandbox.common.exception.InvalidDatasetException;
 import eu.europeana.metis.sandbox.common.exception.ServiceException;
-import eu.europeana.metis.sandbox.dto.report.DatasetInfoDto;
+import eu.europeana.metis.sandbox.common.locale.Country;
+import eu.europeana.metis.sandbox.common.locale.Language;
```
jochen-vermeulen, line 15: Unused imports :)

src/main/java/eu/europeana/metis/sandbox/entity/DatasetEntity.java
```diff
@@ -26,9 +28,15 @@
   @Column(insertable = false, updatable = false)
   private LocalDateTime createdDate;
 
-  public DatasetEntity(String datasetName, Integer recordsQuantity) {
+  private Language language;
+
+  private Country country;
```
jochen-vermeulen, line 33: 1. Did you check the actual values in the database? Is it using the right values in the columns? There are two modes, one is bu enum name and the other by enum number (i.e. position in the list). We always want the first, but I don't remember what the default is.
2. Could you please put a very clear comment in the two enum classes indicating that the names are used as database keys and should therefore not be changed without taking existing database records into account?

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetReportServiceImpl.java
JoanaCMS: Weird, I thought I pushed the removal of these imports. Oh well! Removed :)

src/main/java/eu/europeana/metis/sandbox/entity/DatasetEntity.java
JoanaCMS: 1. I just checked and indeed they saved the `ordinal()` values in the database. I just fixed that.
2. Just added a comment in each enum class with that warning

PR #32, opened 2021-11-04
Feat/met 3926 b file upload handling

stzanakis:


src/test/java/eu/europeana/metis/sandbox/service/workflow/HarvestServiceImplTest.java
```diff
@@ -0,0 +1,86 @@
...
+import org.springframework.web.multipart.MultipartFile;
+
+
+@ExtendWith(MockitoExtension.class)
+public class HarvestServiceImplTest {
+
+
+  private final ZipFileReader zipfilereader = new ZipFileReader();
+
+  @InjectMocks
```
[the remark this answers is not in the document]
adolfopeixinho: Fixed

src/main/java/eu/europeana/metis/sandbox/service/workflow/HarvestServiceImpl.java
```diff
@@ -0,0 +1,97 @@
+package eu.europeana.metis.sandbox.service.workflow;
```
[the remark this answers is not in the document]
adolfopeixinho: Resolved after our latest chat

src/test/java/eu/europeana/metis/sandbox/service/workflow/HarvestServiceImplTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+    var records = harvestService.harvest(dataset);
+    // In spite records being the same size, they are not in the same order.
+    // So comparing with assertArrayEquals() will output a failed assertion
+    // assertArrayEquals(expectedRecords.toArray(),records.toArray());
+    assertEquals(expectedRecords.size(), records.size());
+  }
+
+  @Test
+  void harvestService_expectFailNonExistingFile() {
```
stzanakis, line 63: This seems to be a check for the url not the file as described in the name of the method.
Also it actually fails with indeed a `ServiceException` but I think it fails because the String passed to it is not in a URL format and that it's a missing file.
A check we need to have, but the way it's handled is misleading.

src/test/java/eu/europeana/metis/sandbox/service/workflow/HarvestServiceImplTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+    Path dataSetPath = Paths.get("src", "test", "resources", "zip", "non-existing-file.zip");
+
+    assertFalse(Files.exists(dataSetPath));
+
+    assertThrows(ServiceException.class, () -> harvestService.harvest(dataSetPath.toString()));
+  }
+
+  @Test
+  void harvestService_expectFailCorruptFile() {
```
stzanakis, line 73: Same with the previous method here.

src/test/java/eu/europeana/metis/sandbox/service/workflow/HarvestServiceImplTest.java
```diff
@@ -0,0 +1,82 @@
...
+    var records = harvestService.harvest(dataSetPath.toUri().toString());
+
+    // In spite records being the same size, they are not in the same order.
+    // So comparing with assertArrayEquals() will output a failed assertion
+    // assertArrayEquals(expectedRecords.toArray(),records.toArray());
+    assertEquals(expectedRecords.size(), records.size());
+  }
+
+  @Test
+  void harvestServiceFromUploadFile_expectSuccess() throws IOException {
```
stzanakis, line 44: Perhaps add failure tests for this case as well?

src/test/java/eu/europeana/metis/sandbox/service/workflow/HarvestServiceImplTest.java
```diff
@@ -0,0 +1,82 @@
...
+  void harvestServiceFromURL_expectSuccess() throws IOException {
+
+    Path dataSetPath = Paths.get("src", "test", "resources", "zip", "dataset-valid.zip");
+    assertTrue(Files.exists(dataSetPath));
+
+    var expectedRecords = zipfilereader.getContentFromZipFile(Files.newInputStream(dataSetPath));
+
+    var records = harvestService.harvest(dataSetPath.toUri().toString());
+
+    // In spite records being the same size, they are not in the same order.
```
stzanakis, line 37: I'm assuming you wanted to test the size of list of bytes for each record?
If you do want that, one way to do this could be perhaps creating a set of all records sizes for both the expected and retrieved, and the items of those should be identical.
(order is not important in this case)

adolfopeixinho: Correct. It's now implemented

src/test/java/eu/europeana/metis/sandbox/service/workflow/HarvestServiceImplTest.java
adolfopeixinho: Resolved

PR #34, opened 2021-11-04
Feat/met 3983 b oai mph harvest

adolfopeixinho:


src/main/java/eu/europeana/metis/sandbox/controller/DatasetController.java
```diff
@@ -38,69 +39,117 @@
...
+  @ApiOperation("Process the given dataset by HTTP providing a file")
+  @ApiResponses({
+      @ApiResponse(code = 202, message = MESSAGE_FOR_PROCESS_DATASET, response = Object.class)
+  })
+  @PostMapping(value = "dataset/{name}/harvestByFile", produces = APPLICATION_JSON_VALUE)
+  @ResponseStatus(HttpStatus.ACCEPTED)
+  public DatasetIdDto harvestDatasetFromFile(
+      @ApiParam(value = "name of the dataset", required = true) @PathVariable(value = "name") String datasetName,
+      @ApiParam(value = "country of the dataset", required = true, defaultValue = "Netherlands") @RequestParam Country country,
+      @ApiParam(value = "language of the dataset", required = true, defaultValue = "nl") @RequestParam Language language,
```
JoanaCMS, line 82: Don't forget to change the default value for Language

src/main/java/eu/europeana/metis/sandbox/controller/DatasetController.java
```diff
@@ -38,69 +39,117 @@
...
+      @ApiParam(value = "dataset records uploaded in a zip file", required = true) @RequestParam MultipartFile dataset) {
+    checkArgument(namePattern.matcher(datasetName).matches(),
+        "dataset name can only include letters, numbers, _ or - characters");
+
+    List<ByteArrayInputStream> records = harvestService.harvest(dataset);
+
+    checkArgument(records.size() < maxRecords,
+        "Amount of records can not be more than " + maxRecords);
```
JoanaCMS, line 90: The checkArguments done here are the same or very as the other API calls. Could you work around this to avoid code duplication?
You can, for example, create a private method to check the input values and each API call can use it

src/main/java/eu/europeana/metis/sandbox/service/workflow/HarvestService.java
```diff
@@ -0,0 +1,43 @@
...
+import eu.europeana.metis.sandbox.common.exception.ServiceException;
+import java.io.ByteArrayInputStream;
+import java.util.List;
+import org.springframework.web.multipart.MultipartFile;
+
+
+public interface HarvestService {
+
+  /**
+   * Harvest the given file {@link MultipartFile} to a list of byte[], one string per file in the
```
JoanaCMS, line 12: Every javadoc in here have the same description (with the exception of the variables descriptions), could you change them to be more specific?

src/test/java/eu/europeana/metis/sandbox/controller/DatasetControllerIT.java
```diff
@@ -27,18 +32,32 @@
   private MockMvc mvc;
```
JoanaCMS, line 33: There are more calls to test here

src/test/java/eu/europeana/metis/sandbox/controller/DatasetControllerTest.java
```diff
@@ -49,16 +53,18 @@
...
+  private HarvestService harvestService;
 
   @MockBean
   private DatasetService datasetService;
 
   @MockBean
   private DatasetReportService datasetReportService;
 
+  private final TestUtils testUtils = new TestUtils();
+
```
JoanaCMS, line 65: What about tests for OAI-PMH?

src/test/java/eu/europeana/metis/sandbox/service/workflow/HarvestServiceImplTest.java
```diff
@@ -0,0 +1,82 @@
...
+import java.nio.file.Paths;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.springframework.mock.web.MockMultipartFile;
+
+
+@ExtendWith(MockitoExtension.class)
+public class HarvestServiceImplTest {
+
```
JoanaCMS, line 22: What about tests for OAI-PMH harvesting?

src/main/java/eu/europeana/metis/sandbox/controller/DatasetController.java
adolfopeixinho: Done, changed accordingly.

src/main/java/eu/europeana/metis/sandbox/controller/DatasetController.java
adolfopeixinho: I'll just leave this as it is. If you want to discuss this offline. Let me know

src/test/java/eu/europeana/metis/sandbox/controller/DatasetControllerIT.java
adolfopeixinho: correct, no more calls

src/main/java/eu/europeana/metis/sandbox/service/workflow/HarvestService.java
adolfopeixinho: More specific documentation of the harvester library is available in the metis-framework harvester class javadoc. If you believe it is beneficial to add information from there, here, sure. I believe the current javadoc is enough

src/main/java/eu/europeana/metis/sandbox/controller/DatasetController.java
JoanaCMS: It's always preferable to improve the quality of our code, and I believe that it is possible to include the `checkArguments` and the `harvest` function into one private method, give it a try, just as we mentioned in our call ;) 
If it gets too complicated, then just leave it be.

src/test/java/eu/europeana/metis/sandbox/service/workflow/HarvestServiceImplTest.java
JoanaCMS: There are still some tests missing here related to the OAI-PMH

src/main/java/eu/europeana/metis/sandbox/service/workflow/HarvestService.java
JoanaCMS: The changes done here are what I wanted, so it's fine as it is

src/test/java/eu/europeana/metis/sandbox/controller/DatasetControllerIT.java
JoanaCMS: Just realized, the OAI-PMH test call is not here, you don't want to add that one?

src/test/java/eu/europeana/metis/sandbox/controller/DatasetControllerTest.java
JoanaCMS: Are you planning to add the tests related to the OAI-PMH?

adolfopeixinho: Added one test for OAI-PMH with the outcome of expected success.

src/test/java/eu/europeana/metis/sandbox/service/workflow/HarvestServiceImplTest.java
adolfopeixinho: True. haven't figured out, how to test this here.

src/main/java/eu/europeana/metis/sandbox/service/workflow/HarvestServiceImpl.java
```diff
@@ -46,6 +55,41 @@
...
+
+    List<ByteArrayInputStream> records = new ArrayList<>();
+    List<Pair<String, Exception>> exceptions = new ArrayList<>();
+
+    try {
+      OaiRecordHeaderIterator recordHeaderIterator = harvesterOai
+          .harvestRecordHeaders(new OaiHarvest(endpoint, prefix, setSpec));
+      OaiRepository oaiRepo = new OaiRepository(endpoint, prefix);
+
+      recordHeaderIterator.forEach(rh -> {
```
stzanakis, line 69: I do not see any record number limitation, this can stay here for a very long time..

src/test/java/eu/europeana/metis/sandbox/service/workflow/HarvestServiceImplTest.java
stzanakis: True! Why is it difficult?
The harvesterOai calls can be simply mocked.

src/main/java/eu/europeana/metis/sandbox/service/workflow/HarvestService.java
```diff
@@ -26,4 +26,17 @@
...
+  /**
+   * Harvest the given endpoint {@link String} to a list of byte[], one string per file in the zip
+   *
+   * @param endpoint    for zip file containing one or more records
+   * @param setSpec     record specification
+   * @param prefix      record prefix
+   * @return List of byte[]
+   * @throws ServiceException if file is not valid, error reading file, if records are empty
+   */
+  List<ByteArrayInputStream> harvest(String endpoint, String setSpec, String prefix) throws ServiceException;
```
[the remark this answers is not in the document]
JoanaCMS: I'm actually ok either way in this case. You could see these methods as similar to each other, as in, they all do the same thing in the end, but each performs with a different approach. It also shows different type of supports (the input parameters I mean) when it comes to harvesting.
However, I also see the benefits of being explicit with the type of harvesting in the methods names. It helps us distinguish each other and it makes things easier when we're looking for a specific approach.
But in the end, it's just a matter on what everyone agrees :)

src/main/java/eu/europeana/metis/sandbox/service/workflow/HarvestServiceImpl.java
JoanaCMS: Maybe this was the issue you were having before Adolfo?

src/main/java/eu/europeana/metis/sandbox/service/workflow/HarvestService.java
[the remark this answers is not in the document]
stzanakis: Actually they do totally different things.
Furthermore one has `harvest(String endpoint, String setSpec, String prefix)` and the other one `harvest(String url)`..
An endpoint is also a url..
Even the javadoc is not clear in fact, one says `Harvest the given URL` and the other one `Harvest the given endpoint`.
What is this endpoint, what are those parameters for this endpoint? Is this new method the same with the url one but with added parameters? I can go on and on here..
And now that I see the javadoc better, it's even saying `@param endpoint    for zip file containing one or more records`

It's definitely not clear if we give this contract to someone who is not the implementer of this code(and none of us since we know what we are doing here).
So for sure the javadoc has to be way more descriptive of what's going on here but in my opinion the method call should also be named differently.

For example the following is way more clear, I don't even need to read the javadoc:
```
harvestZipMultipartFile(MultipartFile file)
harvestZipUrl(String url)
harvestOaipmhEndpoint(String endpoint, String setSpec, String prefix)
```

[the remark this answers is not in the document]
JoanaCMS: Thanks for pointing this out! I actually thought that each method were similar.
I agree with you then, both the method names and javadoc need improvement and need to be more specific

src/test/java/eu/europeana/metis/sandbox/controller/DatasetControllerIT.java
```diff
@@ -57,6 +58,18 @@ public void processDatasetFromUrl_expectStatus_accepted() throws Exception {
         .andExpect(status().isAccepted());
   }
 
+  @Test
+  public void harvestDatasetWithOAI_PMH_expectStatus_accepted() throws Exception {
+
+    mvc.perform(multipart("/dataset/{name}/harvestOaiPmh", "my-data-set")
+            .param("country", ITALY.xmlValue())
+            .param("language", IT.xmlValue())
+            .param("url", new URI("http://panic.image.ntua.gr:9000/efg/oai").toString())
```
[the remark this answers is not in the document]
stzanakis: We decided to disable this test for now and work it out later.
We can either create another non-integration test with mocking responses or use a repository that we own.

src/main/java/eu/europeana/metis/sandbox/service/workflow/HarvestServiceImpl.java
stzanakis: For this there is a new ticket to handle records limitations better with a message to the user.

src/test/java/eu/europeana/metis/sandbox/service/workflow/HarvestServiceImplTest.java
stzanakis: This comment was marked as resolved..
The class has been renamed to `HttpHarvestServiceImplTest` but it's not that.
The `HarvestServiceImpl` class inside here is still missing tests for the oai functionality.
Anything that we need to know here?

src/test/java/eu/europeana/metis/sandbox/common/TestUtils.java
```diff
@@ -25,4 +36,59 @@ public String readFileToString(String file) throws IOException {
...
+    String zipFolder = UUID.randomUUID().toString();
+    ZipFile zipFile = this.createTempZipFile(providedZipFile, zipFolder);
+
+    List<ByteArrayInputStream> result;
+    try {
+      List<InputStream> streams = this.getContentFromZipFile(zipFile);
+      List<ByteArrayInputStream> tmpResult = new ArrayList<>(streams.size());
+      Iterator<InputStream> streamsIterator = streams.iterator();
+
+      while (true) {
```
JoanaCMS, line 51: Using a `while(true)` loop is a bad code practice. It makes it hard to maintain the code, the only way to break the loop is with a break, and it also gives a risk of an infinite loop. 
Specially in this case, since you're iterating through each element of the `streams` list,  you can find out the size of this list and use it, which makes this `while(true)` loop is completely unnecessary.
One solution using the iterator is to add this condition to the while loop like so:
`while(streamsIterator.hasNext()){...}`
There are other ways to iterate through the list of course, I would personally try to go with a for each loop.
Either way, please change this loop

src/test/java/eu/europeana/metis/sandbox/common/TestUtils.java
```diff
@@ -25,4 +36,59 @@ public String readFileToString(String file) throws IOException {
...
+    File tempFile = File.createTempFile(folder, ".zip");
+    FileUtils.copyInputStreamToFile(content, tempFile);
+    return new ZipFile(tempFile, 5);
+  }
+
+  private List<InputStream> getContentFromZipFile(ZipFile zipFile) throws IOException {
+    List<InputStream> result = new ArrayList<>();
+    Iterator<? extends ZipEntry> entries = zipFile.stream().iterator();
+
+    while (entries.hasNext()) {
```
JoanaCMS, line 81: This `while` loop is a better implementation comparing to the previous one.
Why did you write this `.hasNext()` condition and the previous one as `while(true)`?

JoanaCMS (changes requested):
There are code conflicts, please fix that and have a look into my comments as well

src/test/java/eu/europeana/metis/sandbox/common/TestUtils.java
adolfopeixinho: The main reason for this was I didn't do/write this code. This was a copy/paste from previous ZipFileReader. I was more concerned with doing the right testing. And I agree with you, this code is messy. It was messier, I just removed all the esoteric variables names var1, var2, var3, etc and gave them more descriptive naming. I will follow you considerations

src/main/java/eu/europeana/metis/sandbox/controller/DatasetController.java
```diff
@@ -69,14 +69,6 @@ public DatasetController(HarvestService harvestService,
     this.reportService = reportService;
   }
 
-  /**
```
JoanaCMS, line 72: Why are you removing these Javadocs? It's always good to have javadocs to make it easier for others to understand what each method does. Please do not remove them

src/main/java/eu/europeana/metis/sandbox/controller/DatasetController.java
```diff
@@ -118,23 +101,43 @@ public DatasetIdDto harvestDatasetFromFile(
...
         "Amount of records can not be more than " + maxRecords);
     var datasetObject = datasetService.createDataset(datasetName, country, language, records);
     return new DatasetIdDto(datasetObject);
   }
 
-  /**
-   * GET API call to return the progress status of a given dataset id
-   * @param datasetId The given dataset id to look for
-   * @return The report of the dataset status
-   */
```
JoanaCMS, line 137: In the meantime, add here as well a javadoc for `harvestDatasetOaiPmh`

src/test/java/eu/europeana/metis/sandbox/common/TestUtils.java
JoanaCMS: The `while(true)` loop is still here, please change this loop

src/test/java/eu/europeana/metis/sandbox/common/TestUtils.java
JoanaCMS: I still see no changes in the loop, please do

src/main/java/eu/europeana/metis/sandbox/controller/DatasetController.java
adolfopeixinho: My bad. It was the resolve conflicts. Didn't notice it. I'll fix it

src/main/java/eu/europeana/metis/sandbox/controller/DatasetController.java
adolfopeixinho: My bad. It was the resolve conflicts. Didn't notice it. I'll fix it

src/test/java/eu/europeana/metis/sandbox/common/TestUtils.java
adolfopeixinho: My bad, forgot to unstash changes

PR #35, opened 2021-11-09
MET-4009 Updated CORS mapping configuration

JoanaCMS:


src/main/java/eu/europeana/metis/sandbox/config/web/MvcConfig.java
```diff
@@ -31,7 +30,7 @@ public void addFormatters(FormatterRegistry registry) {
 
   @Override
   public void addCorsMappings(CorsRegistry registry) {
-    if (StringUtils.isNotBlank(corsMapping)) {
+    if (corsMapping.length != 0) {
```
stzanakis, line 33: An empty array probably is allowed in `allowedOrigins` but a null array is not and we are not checking that here.
In fact `corsMapping.length` will fail if the field is `null` before the below `allowedOrigins` execution.
Better use `ArrayUtils.isNotEmpty(array)`

src/main/java/eu/europeana/metis/sandbox/config/web/MvcConfig.java
```diff
@@ -31,7 +30,7 @@ public void addFormatters(FormatterRegistry registry) {
...
       registry.addMapping("/**").allowedOrigins(corsMapping);
```
stzanakis, line 34: Could you please also add the `.allowedMethods("GET", "HEAD", "POST", "PUT", "DELETE", "OPTIONS")`?
The default are only GET,HEAD,POST.

src/main/java/eu/europeana/metis/sandbox/config/web/MvcConfig.java
JoanaCMS: Thanks for sharing this, this is definitely a better option, I will add it

src/main/java/eu/europeana/metis/sandbox/config/web/MvcConfig.java
JoanaCMS: Done

PR #36, opened 2021-11-17
MET-2622 Review Exception Hierarchy

JoanaCMS:


src/main/java/eu/europeana/metis/sandbox/domain/Event.java
```diff
@@ -21,12 +19,8 @@
    * @param recordInfo must not be null
    * @param step       must not be null
    * @param status     must not be null
-   * @throws NullPointerException if any parameter is null
    */
   public Event(RecordInfo recordInfo, Step step, Status status) {
-    requireNonNull(recordInfo, "Record info must not be null");
```
stzanakis, line 27: Well you removed all the checks here and everywhere else.
I don't see the checks occurring somewhere else now.
They might not all be needed depending on how the classes are used, did you check that?

Other than that, this was performed to avoid throwing a non controlled from us exception right?
We could for all those cases wrap them in a try block, catch the exception and wrap it to one of our own.
Same goes for the rest of the changes.
What do you think?

JoanaCMS: I tried to be as careful as possible when removing these checks. I did check how these classes are used, and most are dependent on the input value given in the dataset controller (like creating records), where these checks are already done there. In my point of view, since the values the user can given are already checked, the rest does not need to be checked again, and if there are any issues I consider them as the system or programmers fault. There are other classes used  and created automatically within the system (like creating record logs), those I don't see the user getting access to them in any way, so I also view them as the system's or programmer's fault. If you'd like, we can have a chat offline where I can demonstrate this occasions.

I will have another look to see which scenarios could use the try catch block and throw one of our exceptions, but I believe most are already checked somewhere within the workflow, but I'm not sure.

acerdas84: While I do not have strong feelings towards this, I will like to clarify why I did this in the first place.
Following [Effective Java](https://www.bol.com/nl/nl/p/effective-java/9200000076724633/?Referrer=ADVNLGOO002008M-G-119060242507-S-852446780350-9200000076724633&gclid=Cj0KCQiAkNiMBhCxARIsAIDDKNXXfe0pjrqpYLRCnl9sRlrNdDjv4U8djL6FWDq2yn1xGaKiLXQP4ewaAr7vEALw_wcB) recommendations, from **Chapter 8 Methods. Item 49: Check parameters for validity**:

> You should clearly document all such restrictions and enforce them with checks at the beginning of the method body. This is a special case of the general principle that you should attempt to detect errors as soon as possible after they occur...
> For public and protected methods, use Javadoc @throws tag to document the exception that will be thrown, if a restriction on parameter values is violated...
> The `Objects.requireNonNull method, added in Java 7, is flexible and convenient, so there's no reason to perform null checks manually anymore

So in your argument you say you check how all classes are used and that the checks are already done in other places, but that makes an assumption that those classes will never be used for other purposes (maybe they never will) but if they are then you can not assume that parameters passed to them will be valid, and there comes the error detection as soon as possible. So I did these checks having that in mind and also since this is an open source project, I wanted to be as clear as I could for future developers of the project, so they know the class contracts and what to expect from how they use the classes.
On the other hand if you all agree these checks are too much and you do not see the need of them, I am fine with it, I just wanted to explain why I put them in the first place

stzanakis: Yes yes fail fast!
Thank you for noting the though process behind @acerdas84 !

I have to say though that this might be an overkill for those or maybe any pojos.
We could think about it in a different way though. As you said someone might want to use them(maybe nobody will as you say), but in a different way were some parameters are allowed to be empty(by their logic) and that means that they cannot use this classes because they will always get an error.

As you saw in my comment I tried to perhaps keep them and follow the logic from before.
On the other hand though, it would be fine I think, if we are checking the parameters properly wherever we use those classes(before even setting them, "fail faster") 
Either way, it is not a huge deal here and we can follow any approach.

JoanaCMS: Thank you for your input @acerdas84!

Just like Simon mentioned, I do believe it is a bit of an overkill to add every kind of checks for every POJO. But at the same time we need to be careful and check how each one of them are used. It really depends on each case as well, how we want to use each class, as in can it be null or not, will the user have access to them or not, etc.
In the end, these are all valid points, we just have to reach an agreement on how to handle this.

My main goal when I started was to remove the null check on cases I believe to be an overkill. But now with Simon's suggestion, I'll have another look into these changes I did and see if there is any case where this suggestion would be best to use. In the meantime, if any of you has any strong feelings towards anything, please let me know and we can have an interesting discussion about this :)

stzanakis:
@JoanaCMS One test seems to be failing, please check.

PR #38, opened 2021-12-13
MET-4064 Perform xslt transformation during harvesting

JoanaCMS:


src/main/java/eu/europeana/metis/sandbox/entity/DatasetEntity.java
```diff
@@ -36,11 +36,20 @@
   @Enumerated(EnumType.STRING)
   private Country country;
 
+  private String xsltTransformerEdmExternal = "";
```
stzanakis, line 39: Why initializing here?
Also this is not a transformer but rather the contents of the xslt right?

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetReportServiceImpl.java
```diff
@@ -225,4 +226,8 @@ private void addStepInfo(List<ProgressByStepDto> stepsInfo,
     errorInfoDtoList.sort(Comparator.comparing(x -> x.getRecordIds().get(FIRST)));
     return errorInfoDtoList;
   }
+
+  private boolean isXsltNotEmpty(String xslt){
+    return StringUtils.isNotEmpty(xslt) && StringUtils.isNotBlank(xslt);
```
stzanakis, line 231: You can just do isNotBlank. It checks for null, empty string and only spaces string.

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetServiceImpl.java
```diff
@@ -26,34 +31,53 @@
...
     this.datasetRepository = datasetRepository;
     this.publishService = publishService;
+    this.transformationService = transformationService;
   }
 
   @Transactional
   @Override
   public Dataset createDataset(String datasetName, Country country, Language language,
-      List<ByteArrayInputStream> records) {
+      List<ByteArrayInputStream> records, InputStream xsltTransformerEDMExternal) {
```
stzanakis, line 50: This is not a transformer but the contents of the xslt file and in fact an input stream. Perhaps name accordingly?

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetServiceImpl.java
```diff
@@ -26,34 +31,53 @@
...
       entity = datasetRepository.save(entity);
     } catch (Exception e) {
       throw new ServiceException(format("Error creating dataset: [%s]. ", datasetName), e);
     }
 
-    var datasetId = String.valueOf(entity.getDatasetId());
-    var dataset = generatorService
+    String datasetId = String.valueOf(entity.getDatasetId());
+
+    // Extra transformation step occurs here
```
stzanakis, line 67: This comment whispers "I would like to be a method of my own" :D

src/main/java/eu/europeana/metis/sandbox/service/workflow/TransformationServiceImpl.java
```diff
@@ -41,12 +51,44 @@ public RecordInfo transform(Record record) {
...
+
+    return recordToEdmExternal;
+  }
+
   private XsltTransformer getTransformer(String datasetName, String edmCountry,
       String edmLanguage) {
     return xsltTransformer.getObject(datasetName, edmCountry, edmLanguage);
   }
 
   private String getXmlDatasetName(Record record) {
```
stzanakis, line 87: I would rename those a bit. It can be confusing, we are getting a datasetName but we already have it from the record information. So what do we get from those methods?

src/main/resources/database/schema.sql
```diff
@@ -17,6 +17,7 @@ create table if not exists dataset (
    created_date timestamp with time zone default now(),
    country varchar(35) not null,
    language varchar(3) not null,
+   xslt_transformer_edm_external text,
```
stzanakis, line 20: Same here, it is not a transformer by itself

src/main/java/eu/europeana/metis/sandbox/controller/DatasetController.java
```diff
@@ -217,6 +237,19 @@ public ProgressInfoDto retrieveDataset(
...
+    if(xslt != null && !xslt.isEmpty()){
+      checkArgument(xslt.getContentType().equals("application/xslt+xml"),
+          "The given xslt file should be a single xml file.");
+      try {
+        return new ByteArrayInputStream(xslt.getBytes());
+      } catch (IOException e) {
+        throw new XsltProcessingException("Something wrong happened while processing xslt file.", e);
+      }
+    }
+    return null;
```
stzanakis, line 250: Perhaps ByteArrayInputStream(new byte[0])?

src/main/java/eu/europeana/metis/sandbox/service/workflow/TransformationService.java
```diff
@@ -15,4 +18,18 @@
...
+   * Transform the given record to EDM external format using a xslt given by the user previously
+   *
+   * @param record must not be null
+   * @return {@link RecordInfo} containing record transformed
+   * @throws NullPointerException      if record is null
+   * @throws RecordProcessingException if records fails at transformation
+   */
+  RecordInfo transformToEdmExternal(Record record);
+
+  public byte[] transformToEdmExternal(String datasetId,
```
stzanakis, line 32: public is default and can be omitted

src/main/java/eu/europeana/metis/sandbox/service/workflow/TransformationServiceImpl.java
```diff
@@ -41,12 +51,44 @@ public RecordInfo transform(Record record) {
     return new RecordInfo(Record.from(record, recordTransformed));
   }
 
+  @Override
+  public RecordInfo transformToEdmExternal(Record record) {
```
stzanakis, line 55: This does not seem to be used anywhere?
That means that the DatasetRepository is not required for this class either?

src/main/java/eu/europeana/metis/sandbox/service/workflow/TransformationServiceImpl.java
```diff
@@ -41,12 +51,44 @@ public RecordInfo transform(Record record) {
...
+        record.getDatasetId())).getXsltTransformerEdmExternal();
+
+    return new RecordInfo(Record.from(record, transformToEdmExternal(record.getDatasetId(),
+        record.getDatasetName(),
+        new ByteArrayInputStream(xsltToEdmExternal.getBytes(StandardCharsets.UTF_8)),
+        record.getCountry(), record.getLanguage(), record.getContent())));
+  }
+
+  @Override
+  public byte[] transformToEdmExternal(String datasetId, String datasetName, InputStream xsltToEdmExternal,
```
stzanakis, line 67: Does this method care what we are transforming to?
The input stream can be anything, so this can have a more general name as well as its parameters.
Also the parameters of datasetName, country and language are specific to the transformation to internal and here they are not required.

src/main/java/eu/europeana/metis/sandbox/dto/DatasetInfoDto.java
```diff
@@ -28,14 +26,18 @@
   @JsonProperty("country")
   private final Country country;
 
+  @JsonProperty("xslt-transformed")
+  private final boolean xsltTransformed;
```
stzanakis, line 30: All of the datasets are transformed to edm internal.
Are we descriptive enough here? What does anyone understand by reading this name?

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetService.java
```diff
@@ -26,7 +28,7 @@
    * @see Dataset
    */
   Dataset createDataset(String datasetName, Country country, Language language,
```
stzanakis, line 30: We are breaking the contract here, while we could simply add another descriptive method and internally one would call the other to avoid duplications.
Not very important here because it's internal code, so I'll leave it to your own discretion.

stzanakis (commented):
I did not check the tests, I think they might change slightly if there are review changes.

src/main/java/eu/europeana/metis/sandbox/entity/DatasetEntity.java
JoanaCMS: I don't remember why I did it, I will remove the initialization.
The name is also a good point, I will change that

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetReportServiceImpl.java
JoanaCMS: Good point, I will change this

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetServiceImpl.java
JoanaCMS: Ahahah that's true! I will take care of that ;)

src/main/java/eu/europeana/metis/sandbox/service/workflow/TransformationServiceImpl.java
JoanaCMS: I see your point, I've renamed them, not sure if it's better. Let me know what you think of `getNewXmlName`.
If you come up with another idea I'm all ears

src/main/resources/database/schema.sql
JoanaCMS: Changed the name accordingly to the one in DatasetEntity

src/main/java/eu/europeana/metis/sandbox/controller/DatasetController.java
JoanaCMS: That is definitely better, will change that

src/main/java/eu/europeana/metis/sandbox/service/workflow/TransformationService.java
JoanaCMS: True, will remove it

src/main/java/eu/europeana/metis/sandbox/service/workflow/TransformationServiceImpl.java
JoanaCMS: Indeed, it is not used anywhere, but I've decided to keep this method because I believe this will be needed in a future ticket (MET-4083). If you prefer I can remove it, but I believe this could be helpful for the ticket in mind

src/main/java/eu/europeana/metis/sandbox/dto/DatasetInfoDto.java
JoanaCMS: This field is supposed to indicate if together with the dataset a xslt file was given, that is, if the extra transformation step was performed or not. This is what Andy and I agreed to write in the report, he uses this field to let know the user that the given xslt file went through. If you believe this name is not explicit enough, I can always talk with Andy again about this and change it to a more descriptive name. What do you think?

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetService.java
JoanaCMS: I don't quite understand what you mean, could you elaborate? Perhaps we can even do a quick call to clarify this

src/main/java/eu/europeana/metis/sandbox/service/workflow/TransformationServiceImpl.java
JoanaCMS: That's a good point, I named these as they are since the main intention is for transformation to EDM external. I could even name the method `transform` to `transformToEdmInternal` to be more specific and name this one `transform`, so basically switching up their names, what do you think?

src/main/java/eu/europeana/metis/sandbox/service/workflow/TransformationServiceImpl.java
stzanakis: getJoinDatasetIdDatasetName

src/main/java/eu/europeana/metis/sandbox/service/workflow/TransformationServiceImpl.java
stzanakis: I think it's better to remove this. It's not relevant to the current ticket.
Who knows maybe we won't do the next ticket, or we might do it differently and forget about the code here.
It's simple to re-introduce.

src/main/java/eu/europeana/metis/sandbox/service/workflow/TransformationServiceImpl.java
stzanakis: Not sure about renaming the other one, since that is a generic one as well as we can simply change the xslt provided to the transformer to something else. On the other hand it's quite specific with the parameter and such so yes we could. I'll leave this up to you.

src/main/java/eu/europeana/metis/sandbox/dto/DatasetInfoDto.java
stzanakis: I get that. But that's a verbal contract between you and Andy. I as an outsider when I try to consume the api call or just read the code, will not understand what this means in particular. Especially because we transform twice. At the very least there should be some javadoc explaining this contract, but I'm in favor of having more descriptive names(in addition to the javadoc if needed).

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetService.java
stzanakis: I meant:
Keeping the old:
Dataset createDataset(String datasetName, Country country, Language language,
List<ByteArrayInputStream> records);

Creating a new:
Dataset createDataset(String datasetName, Country country, Language language,
List<ByteArrayInputStream> records, InputStream xsltTransformerEDMExternal);

And the old calls the new with an empty stream for the last parameter for backwards compatibility.
But as I said this is not hugely important since this code only resides in this project and it's not shared. 
So either way it would be fine.

src/main/java/eu/europeana/metis/sandbox/service/workflow/TransformationServiceImpl.java
JoanaCMS: Done

src/main/java/eu/europeana/metis/sandbox/service/workflow/TransformationServiceImpl.java
JoanaCMS: Ok then, I will remove this method and DataRepository

src/main/java/eu/europeana/metis/sandbox/service/workflow/TransformationServiceImpl.java
JoanaCMS: I switched the method names

src/main/java/eu/europeana/metis/sandbox/dto/DatasetInfoDto.java
JoanaCMS: I understand, I'll handle this

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetService.java
JoanaCMS: Ah now I understand. I see your point, and just like you said, for this specific case it's not really necessary since it's not a shared project, nevertheless I will add this, it's always good to improve the code quality whenever possible, and this could be useful in the future

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetService.java
```diff
@@ -28,6 +29,25 @@
...
       List<ByteArrayInputStream> records);
 
+  /**
+   * Creates a dataset id and publishes the given records for further processing
+   *
+   * @param datasetName must not be null
+   * @param country     must not be null
+   * @param language    must not be null
+   * @param records     must not be null
+   * @param xsltTransformerEDMExternal the xslt file as an input stream
```
stzanakis, line 39: Forgot the name change I believe

src/test/java/eu/europeana/metis/sandbox/service/dataset/DatasetServiceImplTest.java
```diff
@@ -55,7 +55,7 @@ void createDataset_expectSuccess() {
     when(datasetRepository.save(any(DatasetEntity.class))).thenReturn(datasetEntity);
     when(generatorService.generate("1", "name", Country.AUSTRIA, Language.BE, records))
         .thenReturn(dataset);
-    var result = service.createDataset("name", Country.AUSTRIA, Language.BE, records);
+    var result = service.createDataset("name", Country.AUSTRIA, Language.BE, records, null);
```
stzanakis, line 58: I think those method calls should be without the null parameter at the end.
Also we probably want to test success and failure of the new method.

src/test/java/eu/europeana/metis/sandbox/service/workflow/TransformationServiceImplTest.java
```diff
@@ -77,7 +77,7 @@ void transform_invalidXml_expectFail() throws IOException, TransformationExcepti
         .thenThrow(new TransformationException(new Exception("Failing here")));
 
     RecordProcessingException exception = assertThrows(RecordProcessingException.class,
-        () -> service.transform(record));
+        () -> service.transformToEdmInternal(record));
```
stzanakis, line 80: Comment not related to this line.
We need to test the new method in this test class as well right?

src/main/java/eu/europeana/metis/sandbox/service/workflow/TransformationServiceImpl.java
```diff
@@ -41,12 +45,33 @@ public RecordInfo transform(Record record) {
     return new RecordInfo(Record.from(record, recordTransformed));
   }
 
+  @Override
+  public byte[] transform(String datasetId, String datasetName, InputStream xsltToEdmExternal,
```
stzanakis, line 49: Didn't we say that the datasetName, country and language are not needed here?

stzanakis (commented):
There are test failures at the moment as well.

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetService.java
JoanaCMS: You are correct, I missed this change, I will updated it

src/main/java/eu/europeana/metis/sandbox/service/workflow/TransformationServiceImpl.java
JoanaCMS: Oops :) Just removed them

src/test/java/eu/europeana/metis/sandbox/service/dataset/DatasetServiceImplTest.java
JoanaCMS: I have removed the `null` parameters.
I have also added 2 new unit tests for the new `createDataset` method. The existing failure cases I believe already cover the new method.

src/test/java/eu/europeana/metis/sandbox/service/workflow/TransformationServiceImplTest.java
JoanaCMS: Added new unit tests for the new method

PR #39, opened 2021-12-17
Feat/met 4008 caching of the default XSLT 

adolfopeixinho:


src/main/java/eu/europeana/metis/sandbox/service/util/XsltUrlUpdateServiceImpl.java
```diff
@@ -0,0 +1,51 @@
...
+        .uri(URI.create(defaultXsltUrl))
+        .build();
+
+    try (final InputStream xsltStream = httpClient.send(httpRequest, BodyHandlers.ofInputStream())
+        .body()) {
+      String transformXslt = new String(xsltStream.readAllBytes(),
+          StandardCharsets.UTF_8);
+      var entity = transformXsltRepository.findByTransformXslt(transformXslt);
+      // if xslt not in DB, save it
+      if (entity != null && entity.isPresent()) {
```
stzanakis, line 42: This says, if it exists in the db then store again what is already present in the database.
If it doesn't exist, it will never save anything.
So this won't do the update successfully for the first or any subsequent runs.


I think this can be simplified by always storing the retrieved value, no need for checkups.
And when there is an error do nothing to the db(the value should be already there from a previous run).

src/main/java/eu/europeana/metis/sandbox/service/workflow/TransformationServiceImpl.java
```diff
@@ -5,24 +5,23 @@
 import eu.europeana.metis.sandbox.common.exception.RecordProcessingException;
 import eu.europeana.metis.sandbox.domain.Record;
 import eu.europeana.metis.sandbox.domain.RecordInfo;
-import eu.europeana.metis.transformation.service.CacheValueSupplier.CacheValueSupplierException;
-import eu.europeana.metis.transformation.service.EuropeanaGeneratedIdsMap;
+import eu.europeana.metis.sandbox.repository.TransformXsltRepository;
 import eu.europeana.metis.transformation.service.EuropeanaIdCreator;
```
stzanakis, line 9: This import has been removed but it's required.

src/test/java/eu/europeana/metis/sandbox/service/workflow/TransformationServiceImplTest.java
```diff
@@ -1,143 +1,93 @@
...
-  }
+    TransformXsltEntity transformXsltEntity = new TransformXsltEntity(transformFile);
+    when(transformXsltRepository.findById(anyInt())).thenReturn(Optional.of(transformXsltEntity));
 
-  @Test
-  void transform_nullXsltContent_expectFail() {
-    assertThrows(RecordProcessingException.class, () -> service.transform("identifier", null, "record".getBytes(
-        StandardCharsets.UTF_8)));
-  }
+    var result = transformationService.transform(inputRecord);
```
stzanakis, line 68: This gives an error probably another method was meant here?

src/test/java/eu/europeana/metis/sandbox/service/workflow/TransformationServiceImplTest.java
```diff
@@ -1,143 +1,93 @@
...
+        .content(input.getBytes(StandardCharsets.UTF_8))
         .recordId("1").build();
 
-    when(objectProvider.getObject(anyString(), anyString(), anyString()))
-        .thenReturn(xsltTransformer);
-    when(xsltTransformer.transformToBytes(any(byte[].class), any(EuropeanaGeneratedIdsMap.class)))
-        .thenThrow(new TransformationException(new Exception("Failing here")));
+    RecordProcessingException exception =
+        assertThrows(RecordProcessingException.class,
+            () -> transformationService.transform(record));
```
stzanakis, line 89: Same here

stzanakis (changes requested):
On my machine the build is not running because of the some of the issue that I commented about.
Check below.
But I think it also has some issue with DatasetControllerIT.
Please check if the build is sucessful.

src/main/java/eu/europeana/metis/sandbox/service/util/XsltUrlUpdateServiceImpl.java
adolfopeixinho: You're correct. It's fixed now

src/test/java/eu/europeana/metis/sandbox/service/workflow/TransformationServiceImplTest.java
adolfopeixinho: Method name changed. It's fixed now

src/main/java/eu/europeana/metis/sandbox/service/workflow/TransformationServiceImpl.java
adolfopeixinho: Fixed

src/main/java/eu/europeana/metis/sandbox/service/util/XsltUrlUpdateServiceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+        .uri(URI.create(defaultXsltUrl))
+        .build();
+
+    try (final InputStream xsltStream = httpClient.send(httpRequest, BodyHandlers.ofInputStream())
+        .body()) {
+      String transformXslt = new String(xsltStream.readAllBytes(),
+          StandardCharsets.UTF_8);
+      var entity = transformXsltRepository.findByTransformXslt(transformXslt);
+
+      if (entity != null && entity.isPresent()) {
```
stzanakis, line 43: If you have it in the database, do you need to save it again?

stzanakis: Also not sure if the id field in the TransformXsltEntity is useful since you are using the uniqueness here by the transformXslt String content.
Which means the xslt entity can grow indefinitely?
I think what we should better do is re-use the returned object of the database by checking if there is one and only just one(without checking it's content on the query), and then we can update the specific entity(now comparing the content with what we have from the http and decide if we need to update that specific entity) with the same id and not create new ones every time.

src/main/java/eu/europeana/metis/sandbox/service/util/XsltUrlUpdateServiceImpl.java
```diff
@@ -0,0 +1,59 @@
...
+
+    try (final InputStream xsltStream = httpClient.send(httpRequest, BodyHandlers.ofInputStream())
+        .body()) {
+      String transformXslt = new String(xsltStream.readAllBytes(),
+          StandardCharsets.UTF_8);
+      var entity = transformXsltRepository.findByTransformXslt(transformXslt);
+
+
+      if (entity != null && entity.isPresent()) {
+        if (!(transformXslt.equals(entity.get().getTransformXslt()))){
```
stzanakis, line 45: This is still incorrect.
That is because the fetching from the database of transformXsltRepository.findByTransformXslt(transformXslt); above!
Imagine this scenario:
First time we are getting from the URL String "A" and since we don't find it we are storing String "A".
Second time we are getting from the URL String "B" and we  are searching with that and since we don't find it we store it.
So essentially we are creating new ones all the time.

The idea is to replace the existent one.
What you are missing here is searching for the right thing, you'd need to search not based on the String received but rather at the first(there should be only one) item from the database and replace that if it's different.

stzanakis (changes requested):

PR #43, opened 2022-01-14
MET-4193 fix for integration test failing because of bean initializat…

adolfopeixinho:
…ion in xsltUpdateService


jeortizquan (commented):
Tests now are passing ok!
just uncomment. from *SandboxApplicationIT*
```@Disabled("Disabled until fixed, the url initialization of the default xslt is not working if not an http!")```
and update the branch with latest changes.

jeortizquan (approved):

PR #47, opened 2022-01-24
MET-4158 Support supplying processing errors.

stzanakis:


JoanaCMS (commented):
After metis-framework PR is merged, we should rebuild this and check if everything is fine

PR #49, opened 2022-01-26
Feat/met 4196 final changes for tier calculation

stzanakis:


JoanaCMS (approved):
Once the PR from metis-framework that this is dependent on is merged and built, we should re-run the checks in here to make sure everything is ok before merging

PR #54, opened 2022-02-24
MET-4052 Move harvest oai to queue

JoanaCMS:


src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetServiceImpl.java
```diff
@@ -11,134 +11,179 @@
...
+    requireNonNull(records, "Records must not be null");
+    checkArgument(!records.isEmpty(), "Records must not be empty");
+
+    DatasetEntity entity = new DatasetEntity(datasetName, records.size(), language, country,
+        recordLimitExceeded);
+    boolean hasXsltTransformerEdmExternal = false;
+
+    if (isInputStreamAvailable(xsltEdmExternalContentStream)) {
+      try {
+        entity.setXsltEdmExternalContent(new String(xsltEdmExternalContentStream.readAllBytes()));
```
stzanakis, line 97: We need to use encoding when reading the bytes.
We use UTF-8.

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetServiceImpl.java
```diff
@@ -11,134 +11,179 @@
...
+      InputStream xsltEdmExternalContentStream) {
+    requireNonNull(datasetName, "Dataset name must not be null");
+    requireNonNull(country, "Country must not be null");
+    requireNonNull(language, "Language must not be null");
+
+    DatasetEntity entity = new DatasetEntity(datasetName, 0, language, country, false);
+
+    if (isInputStreamAvailable(xsltEdmExternalContentStream)) {
+      try {
+        entity.setXsltEdmExternalContent(new String(xsltEdmExternalContentStream.readAllBytes()));
```
stzanakis, line 62: We need to use encoding when reading the bytes.
We use UTF-8.

src/main/java/eu/europeana/metis/sandbox/service/dataset/AsyncDatasetPublishServiceImpl.java
```diff
@@ -3,75 +3,123 @@
...
+    this.amqpTemplate = amqpTemplate;
+    this.createdQueue = createdQueue;
+    this.oaiHarvestedQueue = oaiHarvestedQueue;
+    this.transformationToEdmExternalQueue = transformationToEdmExternalQueue;
+    this.asyncServiceTaskExecutor = asyncServiceTaskExecutor;
+  }
+
+
+  @Override
+  public CompletableFuture<Void> harvestOaiPmh(String datasetName, String datasetId,
```
stzanakis, line 54: I'll try to elaborate my thoughts here, but I think it applies from here onwards.
In this method we are creating a "fake" record, we are then sending to a queue harvest information and not records which the queues are intended for, effectively "hacking" the previous concept.
The HarvestOaiPmhExecutor is now not the same type of executor like everything else.

Those observations bring me to the question.
Do we need an oai queue in the first place? The harvested of records could be asynchronously performed here and be sent one by one to the created/xslt queue?

src/main/java/eu/europeana/metis/sandbox/domain/RecordProcessEvent.java
```diff
@@ -0,0 +1,63 @@
...
+
+import eu.europeana.metis.sandbox.common.OaiHarvestData;
+import eu.europeana.metis.sandbox.common.Status;
+import eu.europeana.metis.sandbox.common.Step;
+import java.util.List;
+
+/**
+ * Event that contains a record and its processing details
+ */
+public class RecordProcessEvent {
```
stzanakis, line 11: While this was an event that applied to every step, it is now polluted with information of oai harvesting specifically that do not apply to other steps.
I think we should decouple this somehow.

src/main/java/eu/europeana/metis/sandbox/common/amqp/RecordMessageConverter.java
```diff
@@ -38,60 +40,69 @@
   protected static final String STATUS = "status";
   protected static final String STEP = "step";
   protected static final String ERRORS = "errors";
+  protected static final String URL = "url";
+  protected static final String SETSPEC = "setspec";
```
stzanakis, line 44: Those three values could be camel case like the rest.

src/main/java/eu/europeana/metis/sandbox/service/dataset/AsyncDatasetPublishServiceImpl.java
JoanaCMS: This is a good observation. The main reason why we created a oai queue was so we wouldn't wait for the harvesting to be done and for it to start on its own in the "background". This way it will ensure that the user will get an answer fast.

My question now is, if we were to follow your suggestion ("The harvested of records could be asynchronously performed here and be sent one by one to the created/xslt queue?") would we be able to provide an answer to the user fast? Or would we have to wait to finish sending all of the record messages to the queues?

src/main/java/eu/europeana/metis/sandbox/domain/RecordProcessEvent.java
JoanaCMS: I completely agree with you, there should be different type of events containing different type of information.
I'm not sure if this is considered to be part of the ticket, as in if we should focus on this, but we could create a separate ticket dedicated to the improvement of the queue messaging. But I guess this depends on the outcome of the discussion brought up in `AsyncDatasetPublishService`.
What do you think?

src/main/java/eu/europeana/metis/sandbox/service/dataset/AsyncDatasetPublishServiceImpl.java
stzanakis: Yes it should should be asynchronous.
Essentially what is done in HarvestOaiPmhExecutor could be done here calling a method from inside the asynchronous executor `CompletableFuture.runAsync()`. Similarly to the rest.
So the reply will be as fast as like that too.

src/main/java/eu/europeana/metis/sandbox/domain/RecordProcessEvent.java
stzanakis: Indeed, probably fixing the asynchronous issue would not require this information here at all.

src/main/java/eu/europeana/metis/sandbox/service/dataset/AsyncDatasetPublishServiceImpl.java
JoanaCMS: I've tried to test this suggestion after implementing this suggestion (or at least something similar to it) and we do *not* get an immediate answer to the user because we end up waiting for the loop (the one you see in HarvestOaiPmhExecutor) to end, even though we are sending sending each harvested record to a queue. I guess this now depends on the preference of how fast the response should be

src/main/java/eu/europeana/metis/sandbox/executor/workflow/CloseExecutor.java
```diff
@@ -20,19 +20,23 @@
   @Value("${sandbox.rabbitmq.routing-key.closed}")
   private String routingKey;
 
+  @Value("${sandbox.dataset.max-size}")
+  private int maxRecords;
```
stzanakis, line 24: This is not used anymore

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetServiceImpl.java
```diff
@@ -11,134 +11,179 @@
...
+  @Transactional
+  public Dataset createDataset(String datasetName, Country country, Language language,
+      List<ByteArrayInputStream> records, boolean recordLimitExceeded) {
+    return createDataset(datasetName, country, language, records, recordLimitExceeded,
+        new ByteArrayInputStream(new byte[0]));
+  }
+
+  @Override
+  @Transactional
+  public String createEmptyDataset(String datasetName, Country country, Language language,
```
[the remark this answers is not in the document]
stzanakis: There wasn't any mitigation on this.
It's not hugely important, but was there a reason that we cannot do the code deduplication?

src/main/java/eu/europeana/metis/sandbox/service/dataset/AsyncDatasetPublishServiceImpl.java
```diff
@@ -3,75 +3,162 @@
...
 class AsyncDatasetPublishServiceImpl implements AsyncDatasetPublishService {
 
-    private static final Logger LOGGER = LoggerFactory
-            .getLogger(AsyncDatasetPublishServiceImpl.class);
-
+  private static final Logger LOGGER = LoggerFactory.getLogger(
+      AsyncDatasetPublishServiceImpl.class);
+
+
+  private int maxRecords;
```
stzanakis, line 42: Can be final?

src/main/java/eu/europeana/metis/sandbox/common/OaiHarvestData.java
```diff
@@ -0,0 +1,39 @@
...
+ * Class to encapsulate the data that is used for oai-pmh harvesting
+ */
+public class OaiHarvestData {
+
+    private final String url;
+    private final String setspec;
+    private final String metadataformat;
+    private final String oaiIdentifier;
+
+    public OaiHarvestData() {
```
stzanakis, line 13: This is not used anymore.

src/main/java/eu/europeana/metis/sandbox/service/dataset/AsyncDatasetPublishServiceImpl.java
```diff
@@ -3,75 +3,162 @@
...
+              .datasetId(datasetId)
+              .content(new byte[0])
+              .build();
+
+      recordHeaderIterator.forEach(recordHeader -> {
+        OaiHarvestData completeOaiHarvestData = new OaiHarvestData(oaiHarvestData.getUrl(),
+                oaiHarvestData.getSetspec(),
+                oaiHarvestData.getMetadataformat(),
+                recordHeader.getOaiIdentifier());
+        currentNumberOfIterations.getAndIncrement();
```
stzanakis, line 99: Is the counter increment and check in the correct place here?
Shouldn't it be after it's actually successfully sent?

src/main/java/eu/europeana/metis/sandbox/service/dataset/AsyncDatasetPublishServiceImpl.java
```diff
@@ -3,75 +3,162 @@
...
+
+        if (currentNumberOfIterations.get() > maxRecords) {
+          datasetService.updateRecordsLimitExceededToTrue(datasetId);
+          currentNumberOfIterations.set(maxRecords);
```
stzanakis, line 103: And if the counting is after, I think reseting the value could be avoided if the check it >=

stzanakis (changes requested):

src/main/java/eu/europeana/metis/sandbox/service/dataset/AsyncDatasetPublishServiceImpl.java
JoanaCMS: Yup, final it is!

src/main/java/eu/europeana/metis/sandbox/service/dataset/AsyncDatasetPublishServiceImpl.java
stzanakis: I'm pretty sure it's possible, similar approach to any other loop.
But not crucial, I'll resolve the comment.

stzanakis:
I'll merge with squashing in this case

stzanakis:
Has been squash merged

PR #56, opened 2022-03-02
Bug/met 4183 out of memory findings and corrections

jeortizquan:
**Relevant updates:**
- Added unit tests to validate rabbitmq exceptions on executors.
 
- Updated RabbitMQ properties according to the following information.
 spring.rabbitmq.listener.simple.concurrency=2
 spring.rabbitmq.listener.simple.max-concurrency=2 
 
**Parameters based on**:
**Tomcat**: 200 threads max.
**Hikari**: 10 threads x database pool.
**Executor**: 10 threads x publish queue.
**Test machine**: 4 cores, 2 threads per core, 8 threads in parallel. 
**Rabbitmq**: reducing from 10  to 2 concurrent consumers per queue. 22 queues in total. 
Instead of 10x22=220 max concurrent consumers (thinking this is also related with tomcat threads settings, 220>200) to 2x22=44 max concurrent consumers (which is 64 < 200).

### DIFF
--- a/src/main/java/eu/europeana/metis/sandbox/service/util/XsltUrlUpdateServiceImpl.java
+++ b/src/main/java/eu/europeana/metis/sandbox/service/util/XsltUrlUpdateServiceImpl.java
@@ -41,6 +41,7 @@ public class XsltUrlUpdateServiceImpl implements XsltUrlUpdateService {
     } catch (IOException | InterruptedException e) {
       // try to load resource as file
       xsltStream = getClass().getClassLoader().getResourceAsStream(defaultXsltUrl);
+      Thread.currentThread().interrupt();
     } catch (Exception e) {
       LOGGER.warn("Error getting default transform XSLT ", e);
     } finally {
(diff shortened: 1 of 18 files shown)


src/main/java/eu/europeana/metis/sandbox/service/util/XsltUrlUpdateServiceImpl.java
```diff
@@ -31,27 +32,30 @@ public void updateXslt(String defaultXsltUrl) {
...
+                               .GET()
+                               .uri(URI.create(defaultXsltUrl))
+                               .build();
 
       xsltStream = httpClient.send(httpRequest, BodyHandlers.ofInputStream()).body();
 
     } catch (IOException | InterruptedException e) {
       // try to load resource as file
       xsltStream = getClass().getClassLoader().getResourceAsStream(defaultXsltUrl);
+      Thread.currentThread().interrupt();
```
adolfopeixinho, line 44: More a doubt? I this to wait for the stream resource to load?

jeortizquan: This is when ```InterruptedException``` occurs. Sonar reported to include this code line.  This will release the resources of the thread and do not swallow the exception. 
https://stackoverflow.com/questions/4906799/why-invoke-thread-currentthread-interrupt-in-a-catch-interruptexception-block
see the comment of Java Concurrency in practice.

src/test/java/eu/europeana/metis/sandbox/executor/workflow/StepExecutorTest.java
```diff
@@ -0,0 +1,156 @@
...
+import org.mockito.exceptions.base.MockitoException;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.springframework.amqp.AmqpException;
+import org.springframework.amqp.core.AmqpTemplate;
+
+/**
+ * Unit test for {@link StepExecutor}
+ *
+ * @author Jorge Ortiz
+ * @since 01-03-2022
```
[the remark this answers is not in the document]
jeortizquan: oh I see.

PR #59, opened 2022-03-15
MET-4217 Get errors when harvesting fails

JoanaCMS:


src/test/java/eu/europeana/metis/sandbox/service/workflow/HarvestServiceImplTest.java
```diff
@@ -6,10 +6,7 @@
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.ArgumentMatchers.eq;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
jeortizquan, line 9: just leave 
```
import static org.mockito.Mockito.mock;
import static org.mockito.Mockito.spy;
import static org.mockito.Mockito.times;
import static org.mockito.Mockito.verify;
import static org.mockito.Mockito.when;
```
and remove 
```
import static org.mockito.Mockito.*
```
this is why is important: https://rules.sonarsource.com/java/RSPEC-2208

why should we avoid it?

Explicit imports clearly states what external classes the current class is directly using, provided that you don’t leave redundant imports in your code.
If there are two classes with the same name from different packages, it can introduce collision when using wildcard imports.
When multiple people are working in a project, wildcard imports can create confusion as to which classes are actually imported.

jeortizquan (changes requested):
I think is good. there are small changes.

src/test/java/eu/europeana/metis/sandbox/service/workflow/HarvestServiceImplTest.java
```diff
@@ -53,328 +54,390 @@
...
+        Path record2Path = Paths.get("src", "test", "resources", "zip", "Record2.xml");
+        assertTrue(Files.exists(record2Path));
+        List<Path> pathList = new ArrayList<>();
+        pathList.add(record1Path);
+        pathList.add(record2Path);
+
+        return pathList;
+    }
+
+    private Record.RecordBuilder createMockEncapsulatedRecord() throws HarvesterException {
```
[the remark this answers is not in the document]
JoanaCMS: Oops! Fixed

src/test/java/eu/europeana/metis/sandbox/service/workflow/HarvestServiceImplTest.java
JoanaCMS: Thanks for the explanation! I'll update the imports

jeortizquan (approved):
:+1: good

PR #62, opened 2022-03-21
MET-4357 fix resource leak

jeortizquan:
created one instance of MediaExtractor, RdfSerializer, RdfDeserializer

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,12 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
+      - name: Setup ImageMagick
+        run: sudo apt install imagemagick
+      - name: Setup FFmpeg
+        uses: FedericoCarboni/setup-ffmpeg@v1.1.0
+      - name: Setup Ghostscript
+        run: sudo apt-get install ghostscript -y
       - name: Checkout code
         uses: actions/checkout@v2
         with:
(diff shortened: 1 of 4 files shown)


src/main/java/eu/europeana/metis/sandbox/service/workflow/MediaProcessingServiceImpl.java
```diff
@@ -22,75 +19,93 @@
...
+  private final RdfDeserializer rdfDeserializer;
 
   public MediaProcessingServiceImpl(
-      RdfConverterFactory converterFactory,
-      MediaProcessorFactory processorFactory,
-      ThumbnailStoreService thumbnailStoreService) {
-    this.converterFactory = converterFactory;
-    this.processorFactory = processorFactory;
+      ThumbnailStoreService thumbnailStoreService,
+      @Qualifier("rdfSerializer")
```
stzanakis, line 42: The qualifiers are probably not required since they are exactly identical to the name of the field.

.github/workflows/ci.yml
```diff
@@ -6,6 +6,12 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
+      - name: Setup ImageMagick
```
stzanakis, line 9: Is this change required for the unit tests?

jeortizquan: Correct. for the context load Integration test. If not the build fails. It will need to setup this as well in the jenkins build.

src/main/java/eu/europeana/metis/sandbox/service/workflow/MediaProcessingServiceImpl.java
jeortizquan: yes. correct. removed. done!

PR #71, opened 2022-05-04
MET-4499 update api call for get pattern analysis getting latest executionpoint timestamp

jeortizquan:


src/main/java/eu/europeana/metis/sandbox/controller/PatternAnalysisController.java
```diff
@@ -71,13 +74,19 @@ public PatternAnalysisController(PatternAnalysisService<Step, ExecutionPoint> pa
...
-            new DatasetProblemPatternAnalysis<>("0", null, null, new ArrayList<>())));
-
+  public ResponseEntity<DatasetProblemPatternAnalysisView<Step>> getDatasetPatternAnalysis(
+      @ApiParam(value = "id of the dataset", required = true) @PathVariable("id") String datasetId) {
+    Optional<ExecutionPoint> executionPointOptional = executionPointService.getExecutionPoint(datasetId,
+        Step.VALIDATE_INTERNAL.toString());
+    if (executionPointOptional.isPresent()) {
+      return new ResponseEntity<>(new DatasetProblemPatternAnalysisView<>(
+          patternAnalysisService.getDatasetPatternAnalysis(datasetId, Step.VALIDATE_INTERNAL,
+              executionPointOptional.get().getExecutionTimestamp()).orElse(
```
JoanaCMS, line 84: Why do you have this `if...else` block? The `.orElse()` already does this check for you and returns the actual value if is present. What is inside the `orElse()` is the value to return in case **it is not present**.
This `if...else` block is not necessary and it's just repeating behavior `.orElse()` already does for you.
```suggestion
             return new ResponseEntity<>(new DatasetProblemPatternAnalysisView<>(
          patternAnalysisService.getDatasetPatternAnalysis(datasetId, Step.VALIDATE_INTERNAL,
              executionPointOptional.get().getExecutionTimestamp()).orElse(new DatasetProblemPatternAnalysis<>("0", null, null, new ArrayList<>()))), HttpStatus.NOT_FOUND);
```

src/main/java/eu/europeana/metis/sandbox/controller/PatternAnalysisController.java
```diff
@@ -94,13 +103,18 @@ public DatasetProblemPatternAnalysisView<Step> getDatasetPatternAnalysis(
...
-  public List<ProblemPattern> getRecordPatternAnalysis(
+  public ResponseEntity<List<ProblemPattern>> getRecordPatternAnalysis(
       @ApiParam(value = "id of the dataset", required = true) @PathVariable("id") String datasetId,
       @ApiParam(value = "The record content as a file", required = true) @RequestParam String recordId)
       throws SerializationException {
-    String recordContent = recordLogService.getRecordLogEntity(recordId, datasetId, Step.VALIDATE_INTERNAL).getContent();
-    return patternAnalysisService.getRecordPatternAnalysis(rdfConversionUtils.convertStringToRdf(recordContent));
-
+    RecordLogEntity recordLog = recordLogService.getRecordLogEntity(recordId, datasetId, Step.VALIDATE_INTERNAL);
+    if (recordLog != null) {
```
JoanaCMS, line 111: I find these kind of simple `if...else` blocks to be unnecessary. I prefer to use the ternary since it's faster 
```suggestion
    return recordLog != null ? new ResponseEntity<>(patternAnalysisService.getRecordPatternAnalysis(rdfConversionUtils.convertStringToRdf(recordLog.getContent())), HttpStatus.OK) : new ResponseEntity<>(HttpStatus.NOT_FOUND);
```

JoanaCMS (changes requested):

src/main/java/eu/europeana/metis/sandbox/controller/PatternAnalysisController.java
jeortizquan: Let me have a look, is in order to reply the http status when the recordLog is not found, otherwise it throws an unhandled exception.

src/main/java/eu/europeana/metis/sandbox/controller/PatternAnalysisController.java
jeortizquan: I will have a look on rewrite it to return the http status code also when the dataset id is not found. That is the main reason of the ```if...else``` block

jeortizquan: Thanks for the check, actually there was a missing return value in the orElse part so I rewrote this part of the code and now covers two different conditions.
1. If the dataset exists. will proceed to do the analysis else return a not found of the dataset.
2. If the pattern analysis exists. will proceed to process it else return a not found of the analysis.

src/main/java/eu/europeana/metis/sandbox/controller/PatternAnalysisController.java
jeortizquan: rewrote it in positive so is easy to read and added comments.

PR #72, opened 2022-05-06
Debt/met 4483 purge problem pattern data along

JoanaCMS:


src/main/java/eu/europeana/metis/sandbox/repository/RecordErrorLogRepository.java
```diff
@@ -46,6 +46,5 @@ public interface RecordErrorLogRepository extends JpaRepository<RecordErrorLogEn
    * @param datasetId must not be null
    */
   @Modifying
-  @Query("delete from RecordErrorLogEntity where recordId.datasetId = ?1")
-  void deleteByDatasetId(String datasetId);
+  void deleteByRecordIdDatasetId(String datasetId);
```
jeortizquan, line 49: I don't see the relationship between ```record_error_log``` and ```dataset``` tables? does jparepository binds them automatically? does it transforms to this?
``` 
delete from record_error_log rel 
   inner join record re on rel.recordId=re.recordId 
   inner join dataset da on re.datasetid=da.datasetid 
where da.datasetid = 1?
```

JoanaCMS: `record_error_log` has the foreign key `record_id` which is the `id` from `record`.
So this delete query looks into the foreign key and its `datasetId` value
`create table if not exists record_error_log (
    id bigserial,
    record_id bigint not null,
    step varchar(30) not null,
    status varchar(30) not null,
    message text not null,
    stack_trace text not null ,
    created_date timestamp with time zone default now(),
    Primary Key (id),
    Foreign Key (record_id) References record(id)
);

create table if not exists record(
    id bigserial,
    europeana_id varchar(255) null,
    provider_id varchar(255) null,
    dataset_id varchar(100) not null,
    Primary Key (id),
    unique(europeana_id, dataset_id),
    unique(provider_id, dataset_id)
);`

jeortizquan: oh I see thanks.

PR #78, opened 2022-05-31
MET-4362 remove duplicate error reporting for OAIPMH records

jeortizquan:


src/main/java/eu/europeana/metis/sandbox/service/workflow/HarvestServiceImpl.java
```diff
@@ -161,15 +161,9 @@ private RecordInfo harvestOaiRecords(String datasetId, OaiHarvestData oaiHarvest
...
-          oaiHarvestData.getOaiIdentifier(), e);
-      RecordError recordErrorCreated = new RecordError(
-          "Error harvesting OAI-PMH Record Header:" + oaiHarvestData.getOaiIdentifier(),
-          e.getMessage());
-      recordErrors.add(recordErrorCreated);
+      LOGGER.error("Error harvesting OAI-PMH Record Header: {} with exception {}", oaiHarvestData.getOaiIdentifier(), e);
       saveErrorWhileHarvesting(recordToHarvest, oaiHarvestData.getOaiIdentifier(), Step.HARVEST_OAI_PMH, new RuntimeException(e));
-
-      return new RecordInfo(recordToHarvest.build(), recordErrors);
+      return null;
```
JoanaCMS, line 166: Do we know if by returning `null` here will it cause an issue somewhere else in the code? 
There could be a risk of `NullPointerException`

src/main/java/eu/europeana/metis/sandbox/service/workflow/HarvestServiceImpl.java
```diff
@@ -161,15 +161,9 @@ private RecordInfo harvestOaiRecords(String datasetId, OaiHarvestData oaiHarvest
 
       return recordInfo;
     } catch (HarvesterException | IOException e) {
-      LOGGER.error("Error harvesting OAI-PMH Record Header: {} with exception {}",
-          oaiHarvestData.getOaiIdentifier(), e);
-      RecordError recordErrorCreated = new RecordError(
-          "Error harvesting OAI-PMH Record Header:" + oaiHarvestData.getOaiIdentifier(),
-          e.getMessage());
-      recordErrors.add(recordErrorCreated);
```
JoanaCMS, line 169: Could you explain how by removing this solves the problem? I thought that it would be necessary to save the errors for later in the data processing.
Also, since this `recordErrors` list is not being used for saving errors anymore, we can remove it

jeortizquan: The logic was moved inside the method ```saveErrorWhileHarvesting``` and it is stored  in the database.

src/main/java/eu/europeana/metis/sandbox/service/workflow/HarvestServiceImpl.java
jeortizquan: The ```nulls``` are the record with errors that were previously saved on the database. The ```NullPointerException``` is prevented after curating/cleaning the list of records before is transfered to the next step in the workflow.

eu/europeana/metis/sandbox/service/workflow/HarvestServiceImpl.java:132
```  return recordInfoList.stream()
                         .filter(Objects::nonNull)
                         .collect(Collectors.toList());
```
After checking with Jochen found that there is no need to pass faulty record to the next step of the workflow. 

Bug explanation: records with errors where being transfered from harvesting to validate, enrich, transform, publish (propagating the error). Which was causing the counter to be distorted because it was saved twice one in the database and then reprocesed by the queue.

Solution: The record with error is saved in the database and not prevented to be transfered to the next step. For instance if it failed on harvesting (the record stays there), and is not passed to the next step e.g. validate.

PR #80, opened 2022-06-07
MET-4592 Fixed sorting issue of lists within problem patterns objects

JoanaCMS:


src/main/java/eu/europeana/metis/sandbox/controller/PatternAnalysisController.java
```diff
@@ -178,8 +180,22 @@ private DatasetProblemPatternAnalysisView(DatasetProblemPatternAnalysis<T> datas
       this.executionTimestamp = datasetProblemPatternAnalysis.getExecutionTimestamp() == null ? null :
           datasetProblemPatternAnalysis.getExecutionTimestamp().toString();
       this.problemPatternList = datasetProblemPatternAnalysis.getProblemPatternList();
```
jeortizquan, line 182: Actually you can simplify this with the following expression. please check if it does the same as you expected.
```
this.problemPatternList = datasetProblemPatternAnalysis
          .getProblemPatternList()
          .stream()
          .sorted(Comparator.comparing(elem -> elem.getProblemPatternDescription().getProblemPatternId()))
          .collect(Collectors.toList());
```

jeortizquan (commented):
I went through the code, have a look to the comment and see if achieves the same goal.

src/main/java/eu/europeana/metis/sandbox/controller/PatternAnalysisController.java
```diff
@@ -178,8 +180,22 @@ private DatasetProblemPatternAnalysisView(DatasetProblemPatternAnalysis<T> datas
...
+      orderRecordListsInProblemPatternList();
       problemPatternList.sort(Comparator.comparing(elem -> elem.getProblemPatternDescription().getProblemPatternId()));
     }
+
+    private void orderRecordListsInProblemPatternList(){
```
jeortizquan, line 187: You can cleanup this code if the "shorhand" using streams solves the issue.

src/main/java/eu/europeana/metis/sandbox/controller/PatternAnalysisController.java
JoanaCMS: This block of code sorts the problem pattern list itself, which is something I had already done before, which funnily enough is very similar to what I did :) 
This PR is to sort the records in RecordAnalysisList, which is one of the parameter from a ProblemPattern object.
If you have any suggestions for that I would like to hear that!

src/main/java/eu/europeana/metis/sandbox/controller/PatternAnalysisController.java
JoanaCMS: As I mentioned above, the suggested block of code does not do the same thing and this private method.
Would like to hear suggestions if you have any

src/main/java/eu/europeana/metis/sandbox/controller/PatternAnalysisController.java
jeortizquan: Oh I get it now. Inspired in your code we can do the following. This will sort both lists one within the other.
```
   ...
      this.problemPatternList = getSortedProblemPatternList(datasetProblemPatternAnalysis);
    }

    @NotNull
    private List<ProblemPattern> getSortedProblemPatternList(DatasetProblemPatternAnalysis<T> datasetProblemPatternAnalysis) {
      return datasetProblemPatternAnalysis
          .getProblemPatternList()
          .stream()
          .map(currentPattern -> new ProblemPattern(currentPattern.getProblemPatternDescription(),
              currentPattern.getRecordOccurrences(),
              currentPattern.getRecordAnalysisList()
                     .stream()
                     .sorted(Comparator.comparing(RecordAnalysis::getRecordId))
                     .collect(Collectors.toList())))
          .sorted(Comparator.comparing(elem -> elem.getProblemPatternDescription().getProblemPatternId()))
          .collect(Collectors.toList());
    }
```

src/main/java/eu/europeana/metis/sandbox/controller/PatternAnalysisController.java
jeortizquan: see the suggestion.

src/main/java/eu/europeana/metis/sandbox/controller/PatternAnalysisController.java
JoanaCMS: Nice, this suggestion works, thanks! I'll update the code with this

jeortizquan (approved):

PR #82, opened 2022-06-21
MET-4599 changed implementation of XsltUrlUpdateServiceImpl

adolfopeixinho:
…sts accordingly

### DIFF
--- a/src/main/java/eu/europeana/metis/sandbox/repository/TransformXsltRepository.java
+++ b/src/main/java/eu/europeana/metis/sandbox/repository/TransformXsltRepository.java
@@ -1,8 +1,9 @@
 package eu.europeana.metis.sandbox.repository;
 
 import eu.europeana.metis.sandbox.entity.TransformXsltEntity;
+import java.util.Optional;
 import org.springframework.data.jpa.repository.JpaRepository;
 
 public interface TransformXsltRepository extends JpaRepository<TransformXsltEntity, Integer> {
-
+  Optional<TransformXsltEntity> findFirstByIdIsNotNullOrderByIdAsc();
 }
--- a/src/main/java/eu/europeana/metis/sandbox/scheduler/XsltUrlUpdateScheduler.java
+++ b/src/main/java/eu/europeana/metis/sandbox/scheduler/XsltUrlUpdateScheduler.java
@@ -18,7 +18,7 @@ public class XsltUrlUpdateScheduler {
     this.xsltUrlUpdateService = xsltUrlUpdateService;
   }
 
-  @Scheduled(cron = "${sandbox.transformation.xslt-update-frequency}")
+  @Scheduled(cron = "${sandbox.transformation.xslt-update-frequency:0 0 * * * *}")
   public void updateDefaultXsltUrl() {
     xsltUrlUpdateService.updateXslt(defaultXsltUrl);
   }
--- a/src/main/java/eu/europeana/metis/sandbox/scheduler/XsltUrlUpdateScheduler.java
+++ b/src/main/java/eu/europeana/metis/sandbox/scheduler/XsltUrlUpdateScheduler.java
@@ -18,8 +18,7 @@ public class XsltUrlUpdateScheduler {
     this.xsltUrlUpdateService = xsltUrlUpdateService;
   }
 
-  // "0 0 * * * *" = every hour of every day.
-  @Scheduled(cron = "0 0 * * * *")
+  @Scheduled(cron = "${sandbox.transformation.xslt-update-frequency}")
   public void updateDefaultXsltUrl() {
     xsltUrlUpdateService.updateXslt(defaultXsltUrl);
   }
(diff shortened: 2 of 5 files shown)


src/test/java/eu/europeana/metis/sandbox/service/util/XsltUrlUpdateServiceImplTest.java
```diff
@@ -54,66 +54,69 @@ class XsltUrlUpdateServiceImplTest {
...
+    // when
+    doThrow(IllegalArgumentException.class).when(httpClient).sendAsync(any(HttpRequest.class), any());
     setFinalStaticField(XsltUrlUpdateServiceImpl.class, "httpClient", httpClient);
-
-    xsltUrlUpdateService.updateXslt("http://document.domain:12345/xlst");
-
+    xsltUrlUpdateService.updateXslt("http://document.domain:12345/xslt");
+    // then
     Mockito.verify(transformXsltRepository, never()).save(any());
   }
```
jeortizquan, line 118: Is missing a unit test to check with ```responseStatusCode == 404```
```
@Test
  @Order(6)
  void updateXslt_LoadAsFile_ExpectNoUpdate() {
    //given
    wm.stubFor(get("/xslt")
        .withHost(equalTo("document.domain"))
        .withPort(12345)
        .willReturn(notFound()));
    // when
    xsltUrlUpdateService.updateXslt("http://document.domain:12345/xslt");
    // then
    Mockito.verify(transformXsltRepository, never()).save(any());
  }
...
```

src/main/java/eu/europeana/metis/sandbox/service/util/XsltUrlUpdateServiceImpl.java
```diff
@@ -2,81 +2,78 @@
...
 public class XsltUrlUpdateServiceImpl implements XsltUrlUpdateService {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(XsltUrlUpdateServiceImpl.class);
-  private static final HttpClient httpClient = HttpClient.newBuilder().build();
+
   private final TransformXsltRepository transformXsltRepository;
 
+  private static final HttpClient httpClient = HttpClient.newBuilder().version(Version.HTTP_2)
+                                                  .followRedirects(Redirect.NORMAL)
+                                                  .connectTimeout(Duration.ofSeconds(1))
```
jeortizquan, line 30: At least here I would leave a 5 to 30 second timeout what do you think?

src/main/java/eu/europeana/metis/sandbox/service/util/XsltUrlUpdateServiceImpl.java
```diff
@@ -2,81 +2,78 @@
...
-
-  private void closeStream(Closeable closeable) {
-    if (closeable != null) {
-      try {
-        closeable.close();
-      } catch (IOException e) {
-        LOGGER.error("Unable to close stream transform XSLT", e);
+        transformXsltRepository.save(new TransformXsltEntity(newTransformXslt));
       }
+    } catch (DataAccessException e) {
```
jeortizquan, line 75: This exception is not covered also by the tests.

src/main/java/eu/europeana/metis/sandbox/repository/TransformXsltRepository.java
```diff
@@ -1,8 +1,9 @@
 package eu.europeana.metis.sandbox.repository;
 
 import eu.europeana.metis.sandbox.entity.TransformXsltEntity;
+import java.util.Optional;
 import org.springframework.data.jpa.repository.JpaRepository;
 
 public interface TransformXsltRepository extends JpaRepository<TransformXsltEntity, Integer> {
-
+  Optional<TransformXsltEntity> findFirstByIdIsNotNullOrderByIdAsc();
```
jeortizquan, line 8: why ```findFirstByIdIsNotNullOrderByIdAsc```? it is a primary key so never should be ```null``` the value is ```SERIAL``` according to the DB definition. Can you validate that only by ```findFirstByIdOrderByIdAsc``` does the same work? if it does it please replace it.

src/main/java/eu/europeana/metis/sandbox/scheduler/XsltUrlUpdateScheduler.java
```diff
@@ -18,8 +18,7 @@ public XsltUrlUpdateScheduler(XsltUrlUpdateService xsltUrlUpdateService) {
     this.xsltUrlUpdateService = xsltUrlUpdateService;
   }
 
-  // "0 0 * * * *" = every hour of every day.
-  @Scheduled(cron = "0 0 * * * *")
+  @Scheduled(cron = "${sandbox.transformation.xslt-update-frequency}")
   public void updateDefaultXsltUrl() {
```
jeortizquan, line 22: remember to include a comment about this on the ticket, that there is a new property when the release is ready to not forget to include in jenkins.

jeortizquan: I mean in the yml properties, also you can define a default value if we forget.
```${sandbox.transformation.xslt-update-frequency:0 0 * * * *}```

jeortizquan (changes requested):
Have a look on the comments please.

src/main/java/eu/europeana/metis/sandbox/repository/TransformXsltRepository.java
adolfopeixinho: This one I maintained the original JPA query. 
Since the `findFirstByIdOrderByIdAsc` requires an `id` as parameter. 
I wish to get the first that isn't `null` so it works as expected, without a parameter

jeortizquan (approved):

PR #85, opened 2022-06-30
MET-4328 Sandbox give UI control over step display

adolfopeixinho:


jeortizquan (changes requested):
At this moment I **cannot approve** this changes.

I just noticed while doing the code review, there is something to point out: There is a mix of two different concepts here.

**Dataset Status:**
```
  HARVESTING_IDENTIFIERS("harvesting Identifiers"),
  COMPLETED("completed"),
  IN_PROGRESS("in progress");
```
**Record Status:**
```
  SUCCESS("success"),
  FAIL("error"),
  WARN("warning");
```


So this needs more design in my opinion this cannot be merged:
```
  SUCCESS("Success"),
  FAIL("Error"),
  WARN("Warning"),
  HARVESTING_IDENTIFIERS("Harvesting Identifiers"),
  COMPLETED("Completed"),
  IN_PROGRESS("In Progress");
```

But we can capitalise each term to see if the problem solves. The possible solution will be only to do this:
**Dataset Status:**
```
  HARVESTING_IDENTIFIERS("Harvesting Identifiers"),
  COMPLETED("Completed"),
  IN_PROGRESS("In progress");
```
**Record Status:**
```
  SUCCESS("Success"),
  FAIL("Error"),
  WARN("Warning");
```

src/main/java/eu/europeana/metis/sandbox/controller/PatternAnalysisController.java
```diff
@@ -23,14 +23,18 @@
...
 import java.util.Map.Entry;
+import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.stream.Collectors;
-
-import org.jetbrains.annotations.NotNull;
+import javax.validation.constraints.NotNull;
```
jeortizquan, line 37: I think this is the right import ```org.jetbrains.annotations.NotNull;```I don't know why it is changed here.

adolfopeixinho: I believe not. Shouldn't it be a javax.validation.not NULL?!
So our code is dependent Intellij jetbrains, cause it's not a maven dependencyd

jeortizquan: yes. I confirmed  the documentation. https://docs.oracle.com/javaee/7/api/javax/validation/constraints/NotNull.html

jeortizquan (approved):

PR #87, opened 2022-07-01
Feat/met 4620 met 4625 custom metrics

jeortizquan:


src/main/java/eu/europeana/metis/sandbox/service/metrics/MetricsServiceImpl.java
```diff
@@ -0,0 +1,226 @@
...
+
+  private ProgressDatasetEntity findProgressDataset(String datasetId) {
+    return progressDatasetRepository.findByDatasetId(datasetId);
+  }
+
+  private ProgressStepEntity findProgressStep(String datasetId, Step step) {
+    return progressStepRepository.findByDatasetIdAndStep(datasetId, step.value());
+  }
+
+  private void getProgressDatasetMetrics(Map<String, Object> metricsMap) {
```
stzanakis, line 137: This is not really a get is it? It is adding values to a passed parameter, so I would name this differently.  
That being said though, I usually try to avoid modifying/extending the passed parameters and return the result I need.  
The caller then extends(copies all) its own object with the returned results.

src/main/java/eu/europeana/metis/sandbox/service/metrics/MetricsServiceImpl.java
```diff
@@ -0,0 +1,226 @@
...
+    stepMetrics(stepCategory, progressStepEntities, Step.TRANSFORM);
+    stepMetrics(stepCategory, progressStepEntities, Step.VALIDATE_INTERNAL);
+    stepMetrics(stepCategory, progressStepEntities, Step.NORMALIZE);
+    stepMetrics(stepCategory, progressStepEntities, Step.ENRICH);
+    stepMetrics(stepCategory, progressStepEntities, Step.MEDIA_PROCESS);
+    stepMetrics(stepCategory, progressStepEntities, Step.PUBLISH);
+    metricsMap.put("MetricsByStep", stepCategory);
+  }
+
+  private void stepMetrics(Map<String, Object> category, List<ProgressStepEntity> progressStepEntities, Step step) {
```
stzanakis, line 179: Some goes here about the modification/extension of the parameter.

src/main/java/eu/europeana/metis/sandbox/service/metrics/MetricsServiceImpl.java
```diff
@@ -0,0 +1,226 @@
...
+    LongSummaryStatistics totalStepStatistics =
```
stzanakis, line 180: Those blocks are identical. 
They can be merged in a function that would accept a Predicate and a ToLongFunction.

src/main/java/eu/europeana/metis/sandbox/service/metrics/MetricsServiceImpl.java
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  private ProgressStepEntity findProgressStep(String datasetId, Step step) {
+    return progressStepRepository.findByDatasetIdAndStep(datasetId, step.value());
+  }
+
+  private void getProgressDatasetMetrics(Map<String, Object> metricsMap) {
+    List<ProgressDatasetEntity> progressDatasetEntities = progressDatasetRepository.findAll();
+
+    LongSummaryStatistics totalStatistics =
```
stzanakis, line 140: Those blocks are identical.  
They can be merged in a function that would accept a Predicate and a ToLongFunction.

src/main/java/eu/europeana/metis/sandbox/service/metrics/DatasetMetrics.java
```diff
@@ -0,0 +1,23 @@
+package eu.europeana.metis.sandbox.service.metrics;
+
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import com.fasterxml.jackson.annotation.JsonAnySetter;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import java.util.Map;
+
+@JsonInclude(Include.NON_EMPTY)
+public class DatasetMetrics {
```
stzanakis, line 10: Why this custom object? Can't we use the MeterRegistry from Micrometer or not?  
Also if we do, I think then the apm agent could potentially pick them up but not sure how that works at this point.  
Check here https://www.elastic.co/guide/en/apm/agent/java/current/metrics.html

stzanakis (changes requested):

src/main/java/eu/europeana/metis/sandbox/service/metrics/MetricsServiceImpl.java
```diff
@@ -0,0 +1,226 @@
...
+  public DatasetMetrics datasetMetrics() {
+    Map<String, Object> metricsMap = new ConcurrentHashMap<>();
+    getProgressDatasetMetrics(metricsMap);
+    getProgressStepMetrics(metricsMap);
+    DatasetMetrics datasetMetrics = new DatasetMetrics();
+    datasetMetrics.setDatasetMetricsMap(metricsMap);
+    return datasetMetrics;
+  }
+
+  @DeleteOperation
```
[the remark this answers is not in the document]
jeortizquan: Is it ok to include this method in the cleanup scheduler that already exists? so after X time the metrics old metrics cleaned up. Sort of the idea of LRU.

src/main/java/eu/europeana/metis/sandbox/service/metrics/MetricsServiceImpl.java
jeortizquan: Refactored.

src/main/java/eu/europeana/metis/sandbox/service/metrics/MetricsServiceImpl.java
jeortizquan: also here refactored.

src/main/java/eu/europeana/metis/sandbox/service/metrics/MetricsServiceImpl.java
jeortizquan: refactored to a function.

src/main/java/eu/europeana/metis/sandbox/service/metrics/MetricsServiceImpl.java
jeortizquan: also here refactored.

src/main/java/eu/europeana/metis/sandbox/service/metrics/DatasetMetrics.java
jeortizquan: Here is used the custom object suggested by Springboot actuator when adding a custom endpoint. So when adding this property the metrics can be exported to elastic.
```
management:
  metrics:
    export:
      elastic:
        host: "https://elastic.example.com:8086"
```

src/main/resources/database/schema_metrics.sql
```diff
@@ -0,0 +1,26 @@
+BEGIN;
```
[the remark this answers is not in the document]
stzanakis: Yes exactly.

src/main/java/eu/europeana/metis/sandbox/executor/EventRecordLogConsumer.java
```diff
@@ -1,27 +1,30 @@
...
+  public EventRecordLogConsumer(RecordLogService recordLogService, MetricsService metricsService) {
     this.recordLogService = recordLogService;
+    this.metricsService = metricsService;
   }
 
   @RabbitListener(queues = "${sandbox.rabbitmq.queues.record.log.queue}", containerFactory = "recordLogFactory",
       autoStartup = "${sandbox.rabbitmq.queues.record.log.auto-start:true}")
   public void logRecord(RecordProcessEvent input) {
     recordLogService.logRecordEvent(input);
+    metricsService.processMetrics(input.getRecord().getDatasetId());
```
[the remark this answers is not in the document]
stzanakis: Metrics have to be accurate in general.
Your concern is the timestamps?
In fact the current implementation only calculates the performance of the whole workflow and not individual steps.
I'd suggest we go for the scheduler, have a faster interval let's say 5-10 seconds and accept that "deviation" of the end timestamp of the dataset, the code will be cleaner in that way and closer to what we want to achieve.
We are planning to have the timestamps for starting/ending soon.

src/main/java/eu/europeana/metis/sandbox/service/metrics/MetricsServiceImpl.java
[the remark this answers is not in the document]
stzanakis: Depends the implementation you'll be using.
There are LRUs that when an item is inserted, it will also check if it goes above the threshold and will automatically remove the lru item, so probably we won't need to do manual cleanups. But do have a good look on the existent implementations.

src/main/java/eu/europeana/metis/sandbox/service/metrics/DatasetMetrics.java
stzanakis: First of all, we don't send anything directly to elasticsearch, it first goes to a proxy, logstash or apm(currently the same machine but different ports).
Okay, but we'll need to make it work as a time series database which I think it might not work with the above(maybe it will just index the documents or so), not sure though.
What I meant is that if we use the objects that the APM understands, then the apm agent we have connected already might pick those information automatically.
We'll have to test those assumptions though, since what I'm saying is just how I understood it from the docs as I haven't performed that in the past.

src/main/java/eu/europeana/metis/sandbox/service/metrics/MetricsServiceImpl.java
```diff
@@ -0,0 +1,223 @@
...
+    stepCategory.putAll(getMetricsFromStep(progressStepEntities, Step.MEDIA_PROCESS));
+    stepCategory.putAll(getMetricsFromStep(progressStepEntities, Step.PUBLISH));
+    progressStepMetrics.put("MetricsByStep", stepCategory);
+    return progressStepMetrics;
+  }
+
+  private Map<String, Object> getMetricsFromStep(List<ProgressStepEntity> progressStepEntities, Step step) {
+    Map<String, Object> category = new HashMap<>();
+    LongSummaryStatistics totalStepStatistics =
+        calcStepStatistics(progressStepEntities, filterStep(step), ProgressStepEntity::getTotal);
```
stzanakis, line 174: yes!

src/main/java/eu/europeana/metis/sandbox/service/metrics/MetricsServiceImpl.java
```diff
@@ -0,0 +1,223 @@
...
+    Map<String, Object> master = new HashMap<>();
+    master.putAll(getStatisticsFromCategory(totalStepStatistics, "TotalRecords"));
+    master.putAll(getStatisticsFromCategory(successStepStatistics, "SuccessRecords"));
+    master.putAll(getStatisticsFromCategory(failStepStatistics, "FailRecords"));
+    master.putAll(getStatisticsFromCategory(warnStepStatistics, "WarnRecords"));
+    category.put(step.value(), master);
+    return category;
+  }
+
+  private LongSummaryStatistics calcStepStatistics(List<ProgressStepEntity> stepEntities,
```
stzanakis, line 194: calculateStepStatistics. It's not that long is it? :)

PR #95, opened 2022-09-01
MET-4718 enrichment step is doesn't stuck anymore because of Jibx exc…

pWoz:
…eption

### DIFF
--- a/src/main/java/eu/europeana/metis/sandbox/service/workflow/EnrichmentServiceImpl.java
+++ b/src/main/java/eu/europeana/metis/sandbox/service/workflow/EnrichmentServiceImpl.java
@@ -34,6 +34,8 @@ class EnrichmentServiceImpl implements EnrichmentService {
     } catch (EnrichmentException|DereferenceException|SerializationException e) {
       result = record.getContent();
       recordErrors.add(new RecordError(new RecordProcessingException(record.getProviderId(), e)));
+    } catch (Exception e) {
+      throw new RecordProcessingException(record.getProviderId(), e);
     }
 
     return new RecordInfo(Record.from(record, result), recordErrors);
--- a/src/main/java/eu/europeana/metis/sandbox/executor/workflow/StepExecutor.java
+++ b/src/main/java/eu/europeana/metis/sandbox/executor/workflow/StepExecutor.java
@@ -37,7 +37,7 @@ class StepExecutor {
       output = new RecordProcessEvent(recordInfo, step, status);
     } catch (RecordProcessingException ex) {
       output = createFailEvent(input, step, ex);
-    } catch (RuntimeException ex) {
+    } catch (Exception ex) {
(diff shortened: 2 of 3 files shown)


src/main/java/eu/europeana/metis/sandbox/service/workflow/EnrichmentServiceImpl.java
```diff
@@ -34,6 +34,8 @@ public RecordInfo enrich(Record record) {
     } catch (EnrichmentException|DereferenceException|SerializationException e) {
       result = record.getContent();
       recordErrors.add(new RecordError(new RecordProcessingException(record.getProviderId(), e)));
+    } catch (Exception e) {
```
JoanaCMS, line 37: As good practice, it is recommended to specify which exception we want to catch. If we leave as it is, then all exceptions are caught, including RunTimeExcpetion types, which it might not be the desired behavior. 
Unless that is the desired behavior? Based on the unit test, it seems so, could you confirm?

pWoz: There are several aspect that should be taken in the consideration here.

1. What is the problem that we want to solve:
Jibx library throws IOException while marshaling RDF to bytes[]. This exception is not catch by sandbox neither in here EnrichmentServiceImpl.enrich nor in here StepExecutor.consume. Sandbox doesn't handle this case (exception) in any way. That is why processing hangs.

2. We cannot simply catch IOException here EnrichmentServiceImpl.enrich or here StepExecutor.consume because this is checked exception and none of the methods involved declares it in the signature (tbh. this is quite mysterious for me why it is happening like this but I assume that there is some sort of 'sneaky throws' used in the Jbix libraries);

3. In my opinion sandbox should be able to handle any exceptional situation. It means that adding dedicated catch clauses (in this case for IOException) will not solve the problem because there maybe another exceptions that we didn't predict that may happen and again processing will hang. So we should have catch clause for the Exception.class where we will handle all unpredicted cases.

4. I agree that catch clause, that I added in EnrichmentServiceImpl.enrich is not placed correctly, because it solves only issues with Enchiment step. I assume that we would like to have this issue solved for all steps.


So, my suggestion is to add catch clause for the Exception.class in the StepExecutor.consume.

What do You think @JoanaCMS ?

JoanaCMS: Hmm I see, that's a good point. We can leave this catch clause as it is then.
My suggestion for  StepExecutor.consume would be instead of adding a new catch clause for Exeception.class, you could change the current catch clause of RunTimeException to Exception. I believe it was put there with that intention of saving the error message in case of **any** exception.

pWoz: Done

PR #97, opened 2022-09-07
MET-4740 Removed path information from temporary record ids

JoanaCMS:


jeortizquan (approved):
Tested with a zip file with duplicates and works as intended.

PR #102, opened 2022-10-04
MET-4657 Updated sandbox with metis changes

JoanaCMS:


src/test/resources/application.yml
```diff
@@ -81,7 +81,10 @@ sandbox:
     xslt-url: https://metis-core-rest-test.eanadev.org/datasets/xslt/default
   enrichment:
     dereference-url: https://metis-dereference-rest-test.eanadev.org
-    enrichment-url: https://metis-enrichment-rest-test.eanadev.org
+    enrichment-properties:
```
jeortizquan, line 84: I am just afraid are this are production credentials? :sweat_smile: if so we need to remove them and use test environments.

jeortizquan (changes requested):

src/test/resources/application.yml
JoanaCMS: Added test info

jeortizquan (approved):

jeortizquan (approved):

PR #103, opened 2022-10-05
MET-4797 Use a HashSet to improve performance in duplicate detection

jeortizquan:


jeortizquan:
We cannot check content provider Id in the step of Harvesting there are cases that would work due records that are not in EDM format.

PR #105, opened 2022-10-20
bug/MET-4845 handling duplicates validate external

jeortizquan:


src/main/java/eu/europeana/metis/sandbox/repository/RecordRepository.java
```diff
@@ -17,8 +17,10 @@ public interface RecordRepository extends JpaRepository<RecordEntity, Long> {
    * @param providerId the provider id value to update with
    */
   @Modifying
-  @Query("UPDATE RecordEntity rec SET rec.europeanaId = ?2, rec.providerId = ?3 WHERE rec.id = ?1")
-  void updateEuropeanaIdAndProviderId(Long recordId, String europeanaId, String providerId);
+  @Query(value ="UPDATE record rec SET europeana_id=?2, provider_id=?3 "
+      + " WHERE rec.id= ?1 "
+      + " AND NOT EXISTS ( SELECT 1 FROM record dc WHERE rec.dataset_id=dc.dataset_id AND (dc.europeana_id=?2 AND dc.provider_id=?3))", nativeQuery = true)
+  int updateEuropeanaIdAndProviderId(Long recordId, String europeanaId, String providerId, String datasetId);
```
JoanaCMS, line 23: Do you still need `datasetId` has one of the parameters? I see that it's not used in the query

src/main/java/eu/europeana/metis/sandbox/service/record/RecordServiceImpl.java
```diff
@@ -27,32 +29,26 @@ public RecordServiceImpl(RecordRepository recordRepository, XmlRecordProcessorSe
...
     final String datasetId = recordToUpdate.getDatasetId();
     final String providerId = xmlRecordProcessorService.getProviderId(recordToUpdate.getContent());
     final String europeanaId = EuropeanaIdCreator.constructEuropeanaIdString(providerId, datasetId);
 
-    RecordEntity recordEntityProviderCheck = recordRepository.findByProviderIdAndDatasetId(providerId, datasetId);
-    RecordEntity recordEntityEuropeanaCheck = recordRepository.findByEuropeanaIdAndDatasetId(europeanaId, datasetId);
-
-    if (recordEntityProviderCheck == null && recordEntityEuropeanaCheck == null) {
+    final int updatedRecords = recordRepository.updateEuropeanaIdAndProviderId(recordToUpdate.getRecordId(), europeanaId, providerId, datasetId);
+    if (updatedRecords == 0) {
```
JoanaCMS, line 39: I believe it's very unlikely, but what if the number of `updatedRecords` is > 1? Is that even possible to happen?
If it does happen, do we throw an error?

PR #111, opened 2022-11-02
debt/MET-4783 Added queue sizes to metrics

JoanaCMS:


src/main/java/eu/europeana/metis/sandbox/config/ServicesConfig.java
```diff
@@ -22,8 +23,10 @@ MetricsService metricsService(
       RecordRepository recordRepository,
       RecordLogRepository recordLogRepository,
       DatasetProblemPatternRepository problemPatternRepository,
-      MeterRegistry meterRegistry) {
-    metricsService = new MetricsServiceImpl(recordRepository, recordLogRepository, problemPatternRepository, meterRegistry);
+      MeterRegistry meterRegistry,
+      AmqpConfiguration amqpConfiguration) {
+    metricsService = new MetricsServiceImpl(recordRepository, recordLogRepository, problemPatternRepository,
+            meterRegistry, amqpConfiguration, amqpConfiguration.getAmqpAdmin());
```
jeortizquan, line 29: I see this object depends `ampqConfiguration` is it possible to remove this parameter amqpConfiguration.getAmqppAdmin and use it from inside the `MetricsServiceImpl`?

jeortizquan (commented):
I have checked the metrics they are now in the actuator, I just have one question if is possible.

src/main/java/eu/europeana/metis/sandbox/config/ServicesConfig.java
JoanaCMS: I left it like that initially for testing purposes, I wanted to inject a mock of AmqpAdmin into `MetricServiceImpl`
But I figured a different way of doing the tests without that extra parameter in the constructor, I'll push the changes

jeortizquan (approved):

PR #113, opened 2022-12-06
feat/MET-4931_MET-4948 do not include message report for problem pattern P7

jeortizquan:
This feature only filters the pattern P7 and removes the extra text in the message report, but the total information is still stored in the database as a reference that is **missing description field**


JoanaCMS (changes requested):
Overall the code looks good, but while testing locally, I got errors saying that tests in class  DatasetControllerIT seem to be failing.
Please fix it

jeortizquan:
> Overall the code looks good, but while testing locally, I got errors saying that tests in class DatasetControllerIT seem to be failing. Please fix it

The test is updated.

PR #114, opened 2022-12-08
MET-4904 Make setting record id safe for multiple nodes

JoanaCMS:


jeortizquan (approved):
It works good. Tested with small and big file for harvesting. Removing some unused code and added little documentation.

jeortizquan (approved):

PR #116, opened 2022-12-15
Debt/MET-5010 Sonarqube cleanup

JoanaCMS:


jeortizquan (commented):
It is good, adding one more suggestion by sonarqube

PR #117, opened 2023-01-02
Feat/MET-2018_MET4957 Update sandbox with new feature

JoanaCMS:


src/main/java/eu/europeana/metis/sandbox/service/workflow/EnrichmentServiceImpl.java
```diff
@@ -3,41 +3,64 @@
...
+            if (result.getRecordStatus().equals(ProcessedResult.RecordStatus.STOP)) {
+                Optional<Report> report = reports.stream().filter(rep -> Objects.equals(rep.getMessageType(), Type.ERROR)).findFirst();
+                throw new RecordProcessingException(record.getProviderId(), handleRecordStopException(report));
+            }
+
+            reports.stream().filter(report -> report.getMessageType().equals(Type.WARN))
+                    .forEach(report -> recordErrors.add(new RecordError(new RecordProcessingException(record.getProviderId(),
+                            new ServiceException(handleErrorMessage(report), null)))));
+
+        } catch (Exception e) {
```
jochen-vermeulen, line 50: This will also catch the exception you're (potentially) throwing in the try block and wrap it, which is probably not what we want.

src/main/java/eu/europeana/metis/sandbox/service/workflow/EnrichmentServiceImpl.java
```diff
@@ -3,41 +3,64 @@
...
+            throw new RecordProcessingException(record.getProviderId(), e);
+        }
+
+        return new RecordInfo(Record.from(record, result.getProcessedRecord()), recordErrors);
     }
 
-    return new RecordInfo(Record.from(record, result), recordErrors);
-  }
+    private String handleErrorMessage(Report report) {
```
jochen-vermeulen, line 57: Perhaps a better name could be conceived for this method? E.g. 'createErrorMessage'?

src/main/java/eu/europeana/metis/sandbox/service/workflow/EnrichmentServiceImpl.java
```diff
@@ -3,41 +3,64 @@
...
+        return String.format("%s Value: %s", report.getMessage() ,report.getValue());
+    }
+
+    private Exception handleRecordStopException(Optional<Report> report) {
+        return report.isPresent() ? new ServiceException(handleErrorMessage(report.get()),
+                new Exception("StackTrace: " + report.get().getStackTrace())) :
+                new ServiceException("Something went wrong when requesting report.", null);
```
jochen-vermeulen, line 64: My IDE suggests replacing this with 
```
return report
    .map(value -> new ServiceException(createErrorMessage(value), new Exception("StackTrace: " + value.getStackTrace())))
    .orElseGet(() -> new ServiceException("Something went wrong when requesting report.", null));
```
This to keep benefiting from the functional nature of optionals.
What do you think?

Also: do we need that innermost exception (with the stacktrace as argument)? or should we just pass null, just like we do with the `ServiceException` in the `Type.WARN` case?

src/main/java/eu/europeana/metis/sandbox/service/workflow/EnrichmentServiceImpl.java
```diff
@@ -3,41 +3,64 @@
...
+        List<RecordError> recordErrors = new LinkedList<>();
+        ProcessedResult<byte[]> result;
+        try {
+            result = enrichmentWorker.process(record.getContentInputStream());
+            Set<Report> reports = result.getReport();
+
+            if (result.getRecordStatus().equals(ProcessedResult.RecordStatus.STOP)) {
+                Optional<Report> report = reports.stream().filter(rep -> Objects.equals(rep.getMessageType(), Type.ERROR)).findFirst();
+                throw new RecordProcessingException(record.getProviderId(), handleRecordStopException(report));
+            }
```
jochen-vermeulen, line 44: Wouldn't it be nicer to move all this code, including the if statement, to the `handleRecordStopException` method below? Then you don't have to pass an optional (but just the reports) and you don't need a return value (the method optionally throws an exception).

JoanaCMS: That's a good idea, I'll do that

src/main/java/eu/europeana/metis/sandbox/service/workflow/EnrichmentServiceImpl.java
JoanaCMS: I thought that this could help catching `RunTimeExceptions`, but you're right, it might catch the exceptions I'm throwing. I'll remove it

src/main/java/eu/europeana/metis/sandbox/service/workflow/EnrichmentServiceImpl.java
JoanaCMS: I like your name suggestion more, I'll change it, thanks!

src/main/java/eu/europeana/metis/sandbox/service/workflow/EnrichmentServiceImpl.java
JoanaCMS: My IDE was also suggesting it, but decided to keep what I had originally so it would be easier to understand what is happening. But I've changed to method because of previous comments, so this looks different now.

That's a good point, I was thinking that perhaps it would be useful to save the Stack trace somewhere for future debugging, but I can change it null

PR #119, opened 2023-01-12
Feat/MET-5001_MET-5029 Improve record selection algorithm

JoanaCMS:


jeortizquan (approved):
The functionality works. here is the list of tests done.

- with negative step size shows error message.
- with step size zero shows error message.
- with one record process good.
- with bigger stepsize than dataset size was able to throw the error.
- with normal step size works as intended.
- with the limit of the dataset size, picks the right amount of records.

PR #120, opened 2023-01-26
feat/MET-4506 make problem patterns interactions safe for multiple nodes

jeortizquan:


src/main/java/eu/europeana/metis/sandbox/service/workflow/InternalValidationServiceImpl.java
```diff
@@ -68,19 +71,19 @@ public RecordInfo validate(Record recordToValidate) {
...
-    final Lock lock = datasetIdLocksMap.computeIfAbsent(datasetId, s -> new ReentrantLock());
+    final Lock lock = datasetIdLocksMap.computeIfAbsent(datasetId, s -> lockRegistry.obtain("generateAnalysis_" + datasetId));
     final ExecutionPoint executionPoint;
     try {
       lock.lock();
       LOGGER.debug("Generate analysis: {} lock, Locked", datasetId);
       final LocalDateTime timestamp = datasetIdTimestampMap.computeIfAbsent(datasetId, s -> getLocalDateTime(datasetId));
       //We have to attempt initialization everytime because we don't have an entry point for the start of the step
       executionPoint = patternAnalysisService.initializePatternAnalysisExecution(datasetId, Step.VALIDATE_INTERNAL, timestamp);
+      patternAnalysisService.generateRecordPatternAnalysis(executionPoint, new String(recordContent, StandardCharsets.UTF_8));
```
stzanakis, line 82: FYI:
This was outside of the lock since it was multi-thread safe.
It was kept there as a performance benefit.
Adding it gives more security I suppose but perhaps some performance degredation.
I'm fine with both so we can keep it here for now and see how it behaves.

src/main/resources/database/schema_lockrepository.sql
```diff
@@ -0,0 +1,87 @@
+BEGIN;
```
stzanakis, line 1: Why are those .sql files needed?
I think those sql tables are for other functionality from spring-integration which we will not be using.
I believe according to this:
https://docs.spring.io/spring-integration/reference/html/jdbc.html#jdbc-lock-registry
The only table needed is the:

CREATE TABLE INT_LOCK  (
LOCK_KEY CHAR(36),
REGION VARCHAR(100),
CLIENT_ID CHAR(36),
CREATED_DATE TIMESTAMP NOT NULL,
constraint INT_LOCK_PK primary key (LOCK_KEY, REGION)
);

src/main/java/eu/europeana/metis/sandbox/config/ConcurrencyJdbcConfig.java
```diff
@@ -0,0 +1,24 @@
...
+import org.springframework.integration.jdbc.lock.DefaultLockRepository;
+import org.springframework.integration.jdbc.lock.JdbcLockRegistry;
+import org.springframework.integration.jdbc.lock.LockRepository;
+
+@Configuration
+public class ConcurrencyJdbcConfig {
+
+  @Bean
+  public DefaultLockRepository defaultLockRepository(DataSource dataSource) {
+    DefaultLockRepository defaultLockRepository = new DefaultLockRepository(dataSource);
```
stzanakis, line 15: For the DefaultLockRepository and/or JdbcLockRegistry(Since I don't know where this would apply):
How about the TTL of the lock, or other similar parameters?
Shouldn't we have those as configuration parameters?
I think it can be useful in some situations where we'd need to adjust those.

In metis core we have configuration properties that control some values related to the locks that I think we should have control of.

src/test/java/eu/europeana/metis/sandbox/service/problempatterns/PatternAnalysisServiceImplIT.java
```diff
@@ -90,6 +96,18 @@ public static void dynamicProperties(DynamicPropertyRegistry registry) {
   @TestConfiguration
   static class Config {
 
+    @Bean
```
stzanakis, line 99: I don't think those two beans are required here.

src/main/resources/database/schema_lockrepository_drop.sql
```diff
@@ -0,0 +1,11 @@
+BEGIN;
```
stzanakis, line 1: According to the create .sql, this perhaps would require cleaning as well.

src/main/java/eu/europeana/metis/sandbox/config/ConcurrencyJdbcConfig.java
```diff
@@ -0,0 +1,24 @@
...
+
+import javax.sql.DataSource;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.integration.jdbc.lock.DefaultLockRepository;
+import org.springframework.integration.jdbc.lock.JdbcLockRegistry;
+import org.springframework.integration.jdbc.lock.LockRepository;
+
+@Configuration
+public class ConcurrencyJdbcConfig {
```
stzanakis, line 11: Should we call this LockRepositoryJdbcConfig?
Makes more sense given the name of the .sql files.

stzanakis (changes requested):

src/main/java/eu/europeana/metis/sandbox/config/ConcurrencyJdbcConfig.java
jeortizquan: true. done.

src/main/resources/database/schema_lockrepository_drop.sql
jeortizquan: cleaned up too :smile:

src/main/java/eu/europeana/metis/sandbox/service/workflow/InternalValidationServiceImpl.java
jeortizquan: The comment I wrote disappear :sweat_smile: I am writing it again. While developing I notice that `generateRecordPatternAnalysis` has a race condition in the ```DO UPDATE SET record_occurrences = problem_patterns.dataset_problem_pattern.record_occurrences + ? ``` if one node right after the other attempt to update, we will end up with different amount of occurrences that is why I decided to move it back in.

src/main/java/eu/europeana/metis/sandbox/config/ConcurrencyJdbcConfig.java
jeortizquan: the previous comment disappeared. 
here is again, I have added the following parameters.
```
  @Value("${spring.integration.defaultlockrepository.prefix:integration.int_}")
  private String prefix;
  @Value("${spring.integration.defaultlockrepository.ttl:10000}")
  private int timeToLive;
  @Value("${spring.integration.defaultlockrepository.region:DEFAULT}")
  private String region;
```

src/test/java/eu/europeana/metis/sandbox/service/problempatterns/PatternAnalysisServiceImplIT.java
jeortizquan: removed. done.

src/main/resources/database/schema_lockrepository.sql
```diff
@@ -0,0 +1,16 @@
...
+CREATE SCHEMA integration;
+SET search_path TO integration;
```
stzanakis, line 3: Do you know what the impact exactly is for this?
It seems quite dangerous to me when I read some documentation.
Also it does not seem to be required for the int_lock..

src/main/resources/database/schema_lockrepository.sql
```diff
@@ -0,0 +1,16 @@
...
+CREATE TABLE integration.int_lock
+(
+    lock_key     CHAR(36)     NOT NULL,
+    region       VARCHAR(100) NOT NULL,
+    client_id    CHAR(36),
+    created_date TIMESTAMP    NOT NULL,
+    CONSTRAINT int_lock_pk PRIMARY KEY (lock_key, region)
+);
+
+CREATE SEQUENCE int_message_seq START WITH 1 INCREMENT BY 1 NO CYCLE;
```
stzanakis, line 14: Is this still required?

src/main/resources/database/schema_lockrepository.sql
jeortizquan: well the intention was to create the `sequence` in the integration database. so it was missing at the end of the script 
`SET search_path TO public;` just to leave things as it were. But for now you are right is not needed. Removed.

src/main/resources/database/schema_lockrepository.sql
jeortizquan: removed. is not required.

PR #121, opened 2023-01-27
feat/MET-4970 XsltUrlUpdateService safe multiple nodes

jeortizquan:
add a distributed lock to remove race condition. Merge after MET-4506


src/main/java/eu/europeana/metis/sandbox/service/util/XsltUrlUpdateServiceImpl.java
```diff
@@ -60,7 +66,10 @@ public void updateXslt(String defaultXsltUrl) {
   }
 
   private void saveDefaultXslt(String newTransformXslt) {
+    final Lock lock = lockRegistry.obtain("saveDefaultXslt");
```
stzanakis, line 69: Comment not blocking this current pr:  

Since we are actually overwriting the value of an existent object in the db, I wonder if we should actually be also locking(in relation to this lock) during the retrieval of that value when a dataset goes through transformation. Perhaps there can be a race condition in that case.
Something to look at in the future.

src/test/java/eu/europeana/metis/sandbox/service/util/XsltUrlUpdateServiceImplTest.java
```diff
@@ -30,6 +32,7 @@
 import org.mockito.Mock;
 import org.mockito.Mockito;
 import org.mockito.junit.jupiter.MockitoExtension;
+import org.springframework.integration.support.locks.LockRegistry;
 
 @ExtendWith({MockitoExtension.class})
 @TestMethodOrder(OrderAnnotation.class)
```
stzanakis, line 38: Comment not blocking this current pr:

I just saw that and although it is not related to this pr I wanted to say that I personally try to avoid ordering of unit tests and if I need order I would supply it directly in code.  
Also it seems to me that an order is not really required here since the db repo is in fact mocked.  
Fruit for thought.

stzanakis (approved):

src/test/java/eu/europeana/metis/sandbox/service/util/XsltUrlUpdateServiceImplTest.java
jeortizquan: In a certain moment of the development the order was required because it was using a "local mongo" for this test so it matter the order running the test while changing the db. Maybe writing a big method that sets the order in code instead of using annotations you mean?

stzanakis: Yes. Then each unit test is independent.
And also other orders could also be checked, orders for example that should fail.

jeortizquan: created followup ticket https://europeana.atlassian.net/browse/MET-5084

src/main/java/eu/europeana/metis/sandbox/service/util/XsltUrlUpdateServiceImpl.java
jeortizquan: true, if this is being updated, then one node can pick the wrong xlst template and other can pick the correct one.

jeortizquan: added this ticket to the backlog for following up. https://europeana.atlassian.net/browse/MET-5085

PR #122, opened 2023-03-27
MET-5035_Dataset_wide_errors

marcin-rp:


src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetLogService.java
```diff
@@ -0,0 +1,17 @@
...
+
+import eu.europeana.metis.sandbox.common.Status;
+import eu.europeana.metis.sandbox.dto.report.DatasetLogDto;
+import java.util.List;
+import org.springframework.transaction.annotation.Transactional;
+
+public interface DatasetLogService {
+
+  @Transactional
+  Void log(String datasetId, Status status, String message, Throwable e);
```
JoanaCMS, line 11: Why did you decide to use `Void` instead of the primitive `void`?
If you were to implement with the primitive `void` it would be consistent with other services

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetLogServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+    log.setMessage(message);
+    log.setStackTrace(ExceptionUtils.getStackTrace(e));
+    log.setStatus(status);
+    datasetLogRepository.save(log);
+    return null;
+  }
+
+  @Override
+  @Transactional
+  public Void logException(String datasetId, Throwable e) {
```
JoanaCMS, line 52: please rename from `e` to `exception` so its clearer what it's supposed to be

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetLogServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+
+  @Override
+  public List<DatasetLogDto> getAllLogs(int datasetId) {
+    return datasetLogRepository.findByDatasetDatasetId(datasetId)
+                               .stream()
+                               .map(entity -> new DatasetLogDto(entity.getMessage(), entity.getStatus()))
+                               .collect(Collectors.toList());
+  }
+
+  private Throwable unwrapFromCompletionException(Throwable e) {
```
JoanaCMS, line 68: Rename here as well please

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetLogServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+    if (e instanceof java.util.concurrent.CompletionException && e.getCause() != null) {
+      e = e.getCause();
+    }
+    return e;
+  }
+
+  private static String buildMessageFromExceptionChain(Throwable e) {
```
JoanaCMS, line 75: Here as well

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetLogServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+          exceptionFormChain.getMessage() != null ? exceptionFormChain.getMessage() : exceptionFormChain.getClass().getName();
+      if (!message.equals(lastMessage)) {
+        distinctMessages.add(message);
+      }
+      lastMessage = message;
+    }
+    return distinctMessages.stream().collect(Collectors.joining(" - "));
+  }
+
+  private static String enrichMessageForUnexpectedExceptions(Throwable e, String message) {
```
JoanaCMS, line 89: Same here

src/test/java/eu/europeana/metis/sandbox/service/dataset/DatasetLogServiceImplTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+    verify(datasetLogRepository).save(logCaptor.capture());
+    DatasetLogEntity savedLog = logCaptor.getValue();
+    assertThat(savedLog.getMessage()).contains(MESSAGE_1);
+    assertThat(savedLog.getMessage()).doesNotContain(MESSAGE_2);
+    assertThat(savedLog.getStackTrace()).isNotBlank();
+    assertThat(savedLog.getStatus()).isEqualTo(Status.FAIL);
+    assertThat(savedLog.getDataset()).isEqualTo(dataset);
+  }
+
```
JoanaCMS, line 108: Perhaps another unit test here? 
To check that there are zero logs when dataset processing goes well

JoanaCMS (changes requested):
I've tested locally and saw that when I try to create a dataset with HTTP harvesting with non-existent url `http://wrong-url.com` it does not behave the same as OAI-PMH when it get a non-existent url. Please change add this behavior and check my comments

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetLogService.java
marcin-rp: I done it cause the method need to be compatible with functional interface:  Function<Throwable, ? extends T> used by exceptionally method. I could change it to void but in any place where I pass it to the exceptionally method, I need to use more complicated lambda:
Instead of
.exceptionally(e -> datasetLogService.logException(createdDatasetId, e));
I need to use
.exceptionally(e -> {
                            datasetLogService.logException(createdDatasetId, e);
                            return null;
                         });

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetLogService.java
```diff
@@ -0,0 +1,17 @@
...
+
+  @Transactional
+  Void logException(String datasetId, Throwable e);
+
+  List<DatasetLogDto> getAllLogs(int datasetId);
```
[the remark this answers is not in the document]
marcin-rp: In fact in other places there is different type used on the db entities level (int) and on service level (String)
For example in the DatasetService
So  to be consistent with it, I eventually unified both places in the DatasetLogService to use String

src/test/java/eu/europeana/metis/sandbox/service/dataset/DatasetLogServiceImplTest.java
marcin-rp: To test the case that no logs are saved in case of successful dataset execution, I added tests to the DatasetControllerTest.

jeortizquan (changes requested):
The feature overal looks good. Can you switch as much as possible the `assertThat` assertions `org.assertj.core.api.Assertions` to an equivalent of JUnit5 like assertions `assertEquals` in `org.junit.jupiter.api.Assertions` ? we want to keep one library of assertions at the moment in the sandbox for the tests.

PR #123, opened 2023-03-30
MET-5035_copy_Dataset_wide_errors

marcin-rp:


src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetLogServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+  private static final Logger LOGGER = LoggerFactory.getLogger(DatasetLogServiceImpl.class);
+
+  @Autowired
+  private DatasetRepository datasetRepository;
+  @Autowired
+  private DatasetLogRepository datasetLogRepository;
+
+  @Override
+  @Transactional
+  public Void log(String datasetId, Status status, String message, Throwable e) {
```
JoanaCMS, line 35: Throwable `exception` please :)

PR #128, opened 2023-05-02
Feat/MET-5253_MET-5295 download option for the published record version

jeortizquan:


src/main/java/eu/europeana/metis/sandbox/service/record/RecordLogServiceImpl.java
```diff
@@ -76,6 +74,20 @@ public String getProviderRecordString(String recordId, String datasetId)
...
+        } else {
+            try {
+                steps = Set.of(Step.valueOf(step));
+            } catch (IllegalArgumentException ignored) {
+                steps = Set.of();
+            }
+        }
+        return steps;
+    }
+
```
jochenvermeulen, line 90: Two thoughts.
First: should this not be moved to DatasetController? It is after all somthing to do with translating client input, the Service class shouldn't be bothered with that?
Second: should this not throw a proper 400 (bad request) error if the string is not recognised?

jeortizquan: Done. make sense what you mentioned in 1st and 2nd question. I extracted and moved the logic to the controller.

PR #130, opened 2023-05-23
Feat/MET-5305 Accept .tar and .tar.gz uploads harvesting

JoanaCMS:


jeortizquan (changes requested):
Please add the followup ticket for the metrics in jira, and write a comment in this ticket to update the `enum` please.

I tested locally with this file 
[sandbox.tar.gz](https://github.com/europeana/metis-sandbox/files/11556698/sandbox.tar.gz)

and gives this error. 
```
{
  "status": "FAILED",
  "portal-publish": "",
  "total-records": 0,
  "processed-records": 0,
  "progress-by-step": [],
  "dataset-info": {
    "dataset-id": "54",
    "dataset-name": "local_24052023_1802",
    "creation-date": "2023-05-24T18:07:02.097385",
    "language": "Danish",
    "country": "Denmark",
    "record-limit-exceeded": false,
    "transformed-to-edm-external": false
  },
  "error-type": "Error harvesting records  - Problem extracting archive. - Unexpected record signature: 0x88b1f",
  "dataset-logs": [
    {
      "type": "FAIL",
      "message": "Error harvesting records  - Problem extracting archive. - Unexpected record signature: 0x88b1f"
    }
  ],
  "records-published-successfully": false
}
```

src/main/java/eu/europeana/metis/sandbox/common/Step.java
```diff
@@ -5,7 +5,7 @@
  */
 public enum Step {
   HARVEST_OAI_PMH("harvest OAI-PMH",1),
-  HARVEST_ZIP("harvest zip", 2),
+  HARVEST_FILE("harvest file", 2),
```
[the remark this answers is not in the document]
JoanaCMS: True, and we also need to update the database with this new value, something to write down for the new release

src/main/java/eu/europeana/metis/sandbox/service/workflow/HarvestPublishServiceImpl.java
```diff
@@ -32,46 +33,48 @@ public HarvestPublishServiceImpl(HarvestService harvestService,
...
-                                                            .language(datasetMetadata.getLanguage());
+                .datasetId(datasetMetadata.getDatasetId())
+                .datasetName(datasetMetadata.getDatasetName())
+                .country(datasetMetadata.getCountry())
+                .language(datasetMetadata.getLanguage());
         return CompletableFuture.runAsync(() -> {
             try (InputStream input = new URL(url).openStream()) {
                 harvestService.harvest(input, datasetMetadata.getDatasetId(), recordDataEncapsulated,
-                    datasetMetadata.getStepSize());
+                        datasetMetadata.getStepSize(), CompressedFileExtension.ZIP);
```
jeortizquan, line 60: this also needs to be updated :sweat_smile:  I tested with url
`http://ftp.eanadev.org/uploads/sandbox.tar.gz`

jeortizquan (changes requested):
harvestByURL is not working yet.

src/main/java/eu/europeana/metis/sandbox/service/workflow/HarvestPublishServiceImpl.java
JoanaCMS: It should be working now!

src/main/java/eu/europeana/metis/sandbox/controller/DatasetController.java
```diff
@@ -372,39 +381,87 @@ private InputStream createXsltAsInputStreamIfPresent(MultipartFile xslt) {
...
-
-        /**
-         * Instantiates a new Language view.
-         *
-         * @param language the language
-         */
-        LanguageView(Language language) {
-            this.name = language.name();
-            this.xmlValue = language.xmlValue();
+        private CompressedFileExtension checkIfFileTypeFromFileHarvestIsValid (MultipartFile uploadedFile){
```
jeortizquan, line 411: My suggestion would be
a) to split the method `checkIfFileTypeFromFileHarvestIsValid` in two parts 
```
private void checkArgumentMultipartFile(MultipartFile uploadedFile) {
String fileContentType = uploadedFile.getContentType();
            if (StringUtils.isEmpty(fileContentType) || 
                  !(fileContentType.contains("x-tar") || fileContentType.contains("zip") || fileContentType.contains("gzip"))) {
                throw new InvalidCompressedFileException(new Exception("Content type file is not supported"));
            }
}  
```
and a separated function just for getting the type
```
CompressedFileExtension getFileContentType(MultipartFile uploadedFile) {
           if (uploadedFile.getContentType().contains("gzip")) {
                return CompressedFileExtension.TAR_GZ;
            } else if (uploadedFile.getContentType().contains("zip")) {
                return CompressedFileExtension.ZIP;
            } else if (uploadedFile.getContentType().contains("x-tar")) {
                return CompressedFileExtension.TAR;
}
```
call the method would look like this...
```
 harvestPublishService.runHarvestFileAsync(dataset, datasetMetadata, getFileContentType(dataset))
                .exceptionally(e -> datasetLogService.logException(createdDatasetId, e));

```

jeortizquan: the same idea would apply to the url link validation... let me know what are your thoughts. also another option could be renaming `checkIfFileTypeFromFileHarvestIsValid` to something that says that is doing the two things at the same time.

jeortizquan (commented):
The code works now, but it feels a little bit entangled. have a look on the comments.

src/main/java/eu/europeana/metis/sandbox/service/workflow/HarvestPublishService.java
```diff
@@ -3,40 +3,43 @@
...
-     * @param datasetMetadata the object that encapsulates all data related to the dataset
+     * @param url                     URL for compressed file containing one or more records
+     * @param datasetMetadata         the object that encapsulates all data related to the dataset
+     * @param compressedFileExtension the content type of the file being uploaded
      * @return A HarvestContent object containing the content of the harvest and a boolean indicating
      * if it reached the max number of records
      * @throws ServiceException if error processing URL, if URL timeout, if records are empty
      */
-    CompletableFuture<Void> runHarvestHttpZipAsync(String url, DatasetMetadata datasetMetadata);
+    CompletableFuture<Void> runHarvestHttpZipAsync(String url, DatasetMetadata datasetMetadata, CompressedFileExtension compressedFileExtension);
```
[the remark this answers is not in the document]
JoanaCMS: You're right, I've changed the name

src/test/java/eu/europeana/metis/sandbox/controller/DatasetControllerTest.java
```diff
@@ -56,6 +56,8 @@
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.stream.Stream;
+
```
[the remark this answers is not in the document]
JoanaCMS: Done!

src/main/java/eu/europeana/metis/sandbox/controller/DatasetController.java
JoanaCMS: With your suggestion it seems that we are repeating the step of checking the what content type do we have.
According to your suggestion, we would do it first for checking if it's valid and again later to see which type of `CompressedFileExtension` to return.
I prefer to keep the way I currently have because it avoids this repetition. What I will do is change the name of the methods so they are not misleading

src/main/java/eu/europeana/metis/sandbox/common/Step.java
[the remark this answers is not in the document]
jeortizquan: https://europeana.atlassian.net/browse/MET-5332

src/main/java/eu/europeana/metis/sandbox/controller/DatasetController.java
jeortizquan: :+1:

jeortizquan (approved):

PR #142, opened 2023-11-01
MET-5085 transformation service check xslt update multiple node

jeortizquan:


src/main/java/eu/europeana/metis/sandbox/service/workflow/TransformationServiceImpl.java
```diff
@@ -1,119 +1,131 @@
...
+            } catch (IOException e) {
+                LOGGER.error("Unable to close transform stream", e);
+            }
+        }
+    }
+
+    private XsltTransformer getTransformer(String datasetName, String edmCountry,
+                                           String edmLanguage) throws TransformationException {
+
+        if (lockRepository.isAcquired(TransformXsltRepository.LOCK_NAME_KEY)) {
```
stzanakis, line 101: This is not really necessary here.
The xslt will be added during the first ever deployment of the application.
From then onwards there would be updates of it, so the transformation would get the one existent or the one that just replaced the older one.
So such a check here is redundant and would make records throw exceptions(rarely though), for no real reason.

stzanakis (changes requested):
I noticed that some of your later commit have a long "diff", is there something wrong with the file formats on your machine or we just happen that those files haven't been updated yet?

jeortizquan:
> I noticed that some of your later commit have a long "diff", is there something wrong with the file formats on your machine or we just happen that those files haven't been updated yet?

true. I will check my template in my IDE probably an update reset my configuration.

stzanakis:
As we discussed, closing this pr since the added change is not required.

PR #144, opened 2023-11-08
bug/MET-5057: Fix configuration naming and values inconsistencies

stzanakis:


src/main/java/eu/europeana/metis/sandbox/config/amqp/CreatedQueueConfig.java
```diff
@@ -12,23 +12,23 @@
  * The type Created queue configuration.
  */
 @Configuration
-public class CreatedQueueConfiguration extends QueueConsumerConfiguration {
+public class CreatedQueueConfig extends QueueConsumerConfig {
 
-  @Value("${sandbox.rabbitmq.queues.record.created.consumers:12}")
+  @Value("${sandbox.rabbitmq.queues.record.created.consumers}")
```
jeortizquan, line 17: the default value is a good to have in case of misconfiguration.  `sandbox.rabbitmq.queues.record.created.consumers:12`, can we set it then to two? like `sandbox.rabbitmq.queues.record.created.consumers:2` also I see where the confusion began. 
the name of the property was changed. original in the sample yml was
```
consumers: 12
max-consumers: 12
prefetch: 3
```
and I see it mutated to
```
concurrency: 12
max-concurrency: 12
prefetch: 3
```

jeortizquan: by mistake also was copy pasted the same way to the docker in test environment because in production is not like this.

src/main/java/eu/europeana/metis/sandbox/config/amqp/QueueConsumerConfig.java
```diff
@@ -5,22 +5,22 @@
...
- * every subclass can be customized by using the SimpleRabbitListenerContainerFactory
- * by default 2 consumers, 2 concurrent and 1 prefetch is set if none is defined
- * subclasses may also define their own defaults.
+ * every subclass can be customized by using the SimpleRabbitListenerContainerFactory.
  */
-public abstract class QueueConsumerConfiguration {
+@Configuration
+public abstract class QueueConsumerConfig {
 
-  @Value("${spring.rabbitmq.listener.simple.concurrency:2}")
```
jeortizquan, line 19: the original name of the property is here.
https://docs.spring.io/spring-boot/docs/current/reference/html/application-properties.html
that is why it is named 
```
spring.rabbitmq.listener.simple.concurrency
spring.rabbitmq.listener.simple.max-concurrency
```

jeortizquan: After having this review, it brought to the surface the two different names of properties. Do you agree if we standardise into one name and use this spring value names? 
```
.concurency
.max-concurrency
.prefetch
```
with default values
```
spring.rabbitmq.listener.simple.concurrency:2
spring.rabbitmq.listener.simple.max-concurrency:2
spring.rabbitmq.listener.simple.prefetch:1
```

src/main/java/eu/europeana/metis/sandbox/config/amqp/CreatedQueueConfig.java
```diff
@@ -43,9 +43,9 @@ public CreatedQueueConfiguration(MessageConverter messageConverter) {
   SimpleRabbitListenerContainerFactory externalValidationFactory(
       SimpleRabbitListenerContainerFactoryConfigurer configurer,
       ConnectionFactory connectionFactory) {
-    super.setConcurrentQueueConsumers(concurrentConsumers);
```
jeortizquan, line 46: I see here the mistake. it will change to the parent class, but not for the actual class right?

jeortizquan (changes requested):
- I agree with the shortening on the config class names.
- I suggest to use one standardise naming for the consumers (e.g. concurrency, max-concurrency and prefetch).
- I suggest to update the sample yml file with the correct naming of properties.

src/main/java/eu/europeana/metis/sandbox/config/amqp/QueueConsumerConfig.java
```diff
@@ -100,9 +37,6 @@ protected SimpleRabbitListenerContainerFactory getSimpleRabbitListenerContainerF
       ConnectionFactory connectionFactory) {
     var factory = new SimpleRabbitListenerContainerFactory();
     configurer.configure(factory, connectionFactory);
```
jeortizquan, line 39: I have a suggestion now that I see the code is simplified. Adding here some extra parameters, to reduce code duplication in extended classes.
```
protected SimpleRabbitListenerContainerFactory getSimpleRabbitListenerContainerFactory(
      SimpleRabbitListenerContainerFactoryConfigurer configurer,
      ConnectionFactory connectionFactory, 
      Integer concurrentQueueConsumers,
      Integer maxConcurrentQueueConsumers,
      Integer messagePrefetchCount) {
    var factory = new SimpleRabbitListenerContainerFactory();
    configurer.configure(factory, connectionFactory);
    factory.setConcurrentConsumers(concurrentQueueConsumers);
    factory.setMaxConcurrentConsumers(maxConcurrentQueueConsumers);
    factory.setPrefetchCount(messagePrefetchCount);
    factory.setMessageConverter(messageConverter);
    return factory;
  }
```

src/main/java/eu/europeana/metis/sandbox/config/amqp/RecordLogQueueConfig.java
```diff
@@ -112,9 +112,10 @@ Binding logDlqBinding() {
   SimpleRabbitListenerContainerFactory recordLogFactory(
       SimpleRabbitListenerContainerFactoryConfigurer configurer,
       ConnectionFactory connectionFactory) {
-    setConcurrentQueueConsumers(concurrentConsumers);
-    setMaxConcurrentQueueConsumers(maxConsumers);
-    setMessagePrefetchCount(messagePrefetchCount);
-    return getSimpleRabbitListenerContainerFactory(configurer, connectionFactory);
```
jeortizquan: update this lines with the proposed method. 
from this:
```
   SimpleRabbitListenerContainerFactory simpleRabbitListenerContainerFactory = getSimpleRabbitListenerContainerFactory(configurer, connectionFactory);
    simpleRabbitListenerContainerFactory.setConcurrentConsumers(concurrentConsumers);
    simpleRabbitListenerContainerFactory.setMaxConcurrentConsumers(maxConsumers);
    simpleRabbitListenerContainerFactory.setPrefetchCount(messagePrefetchCount);
    return simpleRabbitListenerContainerFactory;
```
to this:
```
    return getSimpleRabbitListenerContainerFactory(configurer, connectionFactory, concurrentConsumers, maxConsumers, messagePrefetchCount);
```

jeortizquan (commented):
Please have a look on the suggestion.

src/main/java/eu/europeana/metis/sandbox/config/amqp/QueueConsumerConfig.java
stzanakis: Sure I have applied it now.
But keep in mind that I was following the structure from before..

jeortizquan (approved):

PR #146, opened 2023-11-14
MET-5471: Update docker jre to temurin 21

stzanakis:


Dockerfile
jeortizquan, line 1: I tested locally, by itself doesn't run if the properties are not injected right? are we okay by letting this image to be like this?

jeortizquan (approved):

Dockerfile
stzanakis: I've added the run configurations(Dockerfile and spring) so that they are common for everyone.
Also updated docker-compose.(note: the field `memswap_limit` might be reported as not expected by the IDE, but it is incorrect and does work)

jeortizquan (approved):

PR #173, opened 2024-12-12
MET-6248 add solr and s3 local support

jeortizquan:


docker-compose.yml
```diff
@@ -28,14 +28,42 @@ services:
...
       MONGO_INITDB_ROOT_USERNAME: guest
       MONGO_INITDB_ROOT_PASSWORD: guest
     ports:
       - '27017:27017'
+  solr:
+    build:
+      context: docker/solr/
+      dockerfile: Dockerfile
+    container_name: metis-sandbox-solr
+    restart: always
```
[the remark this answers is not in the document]
jeortizquan: is just personal taste, we can remove this property.

docker/solr/solr-schema.sh
```diff
@@ -0,0 +1,168 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
jeortizquan: In the documentation probably I didn't make it clear. This script needs to be executed right after the docker container is created is a config script.

src/main/java/eu/europeana/metis/sandbox/config/S3Config.java
```diff
@@ -33,11 +36,34 @@ class S3Config {
   private String thumbnailsBucket;
 
   @Bean
-  AmazonS3 s3Client() {
+  @ConditionalOnProperty(
+      name="sandbox.s3.local-enabled",
+      havingValue = "true")
+  AmazonS3 s3ClientHttp() {
```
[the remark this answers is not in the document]
jeortizquan: in this case `sandbox.s3.ssl` is about using the ssl. I wrote the property more in the sense to use the local minio container. I tested with `http` and `https` in the example and the amazon s3 bean didn't worked. Also you can give me some idea of what you meant with the last part of the comment, (the link) to read the documentation that you were looking at.

docker-compose.yml
```diff
@@ -28,14 +28,42 @@ services:
...
-    image: mongo:4.2.9
+    image: mongo:6.0.12
     container_name: metis-sandbox-mongo
     environment:
       MONGO_INITDB_DATABASE: metis-sandbox
       MONGO_INITDB_ROOT_USERNAME: guest
       MONGO_INITDB_ROOT_PASSWORD: guest
     ports:
       - '27017:27017'
+  solr:
```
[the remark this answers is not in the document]
jeortizquan: I will include in the documentation that we use the internal Zookeeper that comes bundled with SOLR
afbeelding

docker/solr/solr-schema.sh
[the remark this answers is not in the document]
jeortizquan: done. this script is now executed when creating the docker image of SOLR. and there is no need to execute manually.

docker/solr/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+USER 0
+RUN apt-get update  \
+    && apt-get install git -y  \
+    && apt-get install rsync -y  \
+    && apt-get install curl -y  \
+    && apt-get install nano -y  \
+    && apt-get clean \
+    && git clone https://github.com/europeana/search
+COPY solr-schema.sh /opt/solr/search
+COPY europeana-collection-aliasing-solr6.6.5-0.0.1-SNAPSHOT.jar /opt/solr/contrib/lib/europeana-collection-aliasing-solr6.6.5-0.0.1-SNAPSHOT.jar
```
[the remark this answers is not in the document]
stzanakis: How is it working without this .jar in metis-framework/metis-indexing test containers?

docker-compose.yml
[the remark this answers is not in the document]
stzanakis: Let's create a ticket to investigate the zookeeper part. We need to replicate as much as possible what we use in our dedicated environments.

docker/solr/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
```
stzanakis, line 11: Where is this file supposed to be? I don't have it.

docker-compose.yml
```diff
@@ -28,14 +28,40 @@ services:
...
+      - docker-entrypoint.sh
+      - solr
+      - start
+      - -c
+      - -f
+
+  localstack-s3:
+    image: localstack/localstack:s3-latest
+    environment:
+      - DEBUG=${DEBUG:-0}
```
stzanakis, line 57: ?

docker/solr/Dockerfile
jeortizquan: this goes here `/opt/solr/contrib/lib/` and is used by solr. This I found about it when replicating solr in the local environment, and obtained it from solr 7 server. Infra Team provided it.

docker-compose.yml
jeortizquan: got it from the documentation. https://docs.localstack.cloud/user-guide/aws/s3/
afbeelding

docker/solr/Dockerfile
stzanakis: yeah but do we need to have the file elsewhere? How can someone cloning this project run the docker, I don't have the file.

docker-compose.yml
stzanakis: Perhaps document or plainly remove. This can be added on a per case basis.

docker/solr/Dockerfile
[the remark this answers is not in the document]
jeortizquan: the testcontainers works there because it don't use the solr schema configuration it uses the default configuration, metis needs to be updated as well, to use it. I created a ticket to follow up.
https://europeana.atlassian.net/browse/MET-6438

docker-compose.yml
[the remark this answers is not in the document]
jeortizquan: it was taking to long to replicate the zookeeper in a separate docker image it was not recognized by the solr docker. That is why at the moment both run on the same image just as `test` environment but the differents is that `test` uses an independent zookeeper from the one that comes with solr. Here is the ticket.
https://europeana.atlassian.net/browse/MET-6439

docker/solr/Dockerfile
jeortizquan: When cloning the repository the file is downloaded to the local folder, and when building the image the file is copied to the docker image so you will have it there.

docker-compose.yml
jeortizquan: removed. done.

docker/solr/Dockerfile
stzanakis: You said you obtained if from A server, now you are saying it is in the search repo yet I don't see it.

jeortizquan: done. it should be there now.

PR #178, opened 2025-04-04
MET-6449 Use new quorum type queues and clean deprecated ones

stzanakis:


src/main/java/eu/europeana/metis/sandbox/config/amqp/AmqpConfiguration.java
```diff
@@ -189,7 +189,7 @@ Declarables queues() {
                     .deadLetterRoutingKey(mediaProcessedDlq).build(),
         QueueBuilder.durable(publishedQueue).deadLetterExchange(exchangeDlq)
                     .deadLetterRoutingKey(publishedDlq).build(),
-        QueueBuilder.durable(deBiasReadyQueue).deadLetterExchange(deBiasReadyDlq)
+        QueueBuilder.durable(deBiasReadyQueue).deadLetterExchange(exchangeDlq)
```
jeortizquan, line 192: the queue name has to be different otherwise will mix messages from normal workflow with debias workflow.

jeortizquan (changes requested):
just one queue is pointing to the wrong one.

src/main/java/eu/europeana/metis/sandbox/config/amqp/AmqpConfiguration.java
jeortizquan: fixed bug indeed.

PR #188, opened 2025-09-25
Feat/met 6810 rework detection of duplicates

stzanakis:


src/main/java/eu/europeana/metis/sandbox/batch/reader/AbstractIdentifiersItemReader.java
```diff
@@ -0,0 +1,108 @@
...
+  protected String harvestParameterId;
+  @Value("#{jobParameters['datasetId']}")
+  protected String datasetId;
+  @Value("#{jobParameters['stepSize']}")
+  protected String stepSize;
+
+  protected final HarvestParameterService harvestParameterService;
+  protected final DatasetExecutionSetupService datasetExecutionSetupService;
+  protected final ExecutionRunRepository executionRunRepository;
+  private final Deque<T> identifiers = new ArrayDeque<>();
```
jeortizquan, line 46: how the queue will behave in a multithreaded or multi reader environment? do order matters?

src/main/java/eu/europeana/metis/sandbox/batch/reader/AbstractIdentifiersItemReader.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    ExecutionRecordExternalIdentifier executionRecordExternalIdentifier = new ExecutionRecordExternalIdentifier();
+    executionRecordExternalIdentifier.setExecutionRun(executionRun);
+    executionRecordExternalIdentifier.setExternalRecordId(extractStringIdentifier(identifier));
+    executionRecordExternalIdentifier.setDeleted(isDeleted(identifier));
+    return executionRecordExternalIdentifier;
+  }
+
+  private synchronized T takeIdentifier() {
```
jeortizquan, line 96: I see here is addressed concurrency. just check performance when handling a big volume of records.

src/main/java/eu/europeana/metis/sandbox/batch/repository/ExecutionRecordRepository.java
```diff
@@ -1,86 +1,141 @@
...
    * @param pageable The paging configuration, including page number and size.
-   * @return A paginated list of ExecutionRecord entities matching the criteria.
+   * @return A paginated list of ExecutionRecord entities excluding duplicates matching the criteria.
    */
-  Page<ExecutionRecord> findByIdentifier_DatasetIdAndIdentifier_ExecutionId(String datasetId, String executionId,
-      Pageable pageable);
+  @Query("""
+      SELECT r
+      FROM ExecutionRecord r
+      WHERE r.id IN (
```
jeortizquan, line 32: my suggestion and recommendation is to substitute `IN` for `EXISTS` it always better for performance, when db tends to be big this `IN` later becomes and issue, `EXISTS` behaves better under large amounts of data. consider this in all cases where `IN` is used in this class, consider the `EXPLAIN ANALYSE` see if this comment applies.

src/main/java/eu/europeana/metis/sandbox/service/workflow/harvest/HarvestService.java
```diff
@@ -0,0 +1,58 @@
...
+   * @param harvestTarget the target object from which the record will be harvested. This can be an instance of a custom class or
+   * object defining the data source details.
+   * @param sourceRecordId the unique identifier of the source record to be harvested from the specified target.
+   * @return a {@code HarvestedRecord} object containing the source record ID, harvested record ID, and its raw data content.
+   * @throws HarvestException if an error occurs during the harvesting process, such as issues with accessing the target or
+   * processing the record.
+   */
+  HarvestedRecord harvestRecord(S harvestTarget, String sourceRecordId) throws HarvestException;
+
+  /**
```
jeortizquan, line 50: should be better to place it in an external file? is small though. gave me the sensation doesn't belong here.

src/main/java/eu/europeana/metis/sandbox/service/workflow/harvest/OaiHarvestService.java
```diff
@@ -0,0 +1,162 @@
...
+import org.springframework.stereotype.Service;
+
+/**
+ * Service for harvesting records from an OAI-PMH compliant repository.
+ */
+@Slf4j
+@Service
+public class OaiHarvestService implements HarvestService<OaiRecordHeader, OaiHarvest> {
+
+  private static final int STOP_WATCH_INTERNAL = 10;
```
jeortizquan, line 39: suggestion to add units `STOP_WATCH_INTERNAL_SECONDS` or `STOP_WATCH_SECONDS_INTERNAL`

src/main/java/eu/europeana/metis/sandbox/service/workflow/harvest/FileHarvestTarget.java
```diff
@@ -0,0 +1,18 @@
...
+import lombok.Value;
+
+/**
+ * Represents a target file to be harvested in the system.
+ * <p>
+ * The FileHarvestTarget class encapsulates the metadata and content of a file to be used for harvesting operations. It includes
+ * the file's name, type, and raw byte content.
+ */
+@Value
+public final class FileHarvestTarget {
```
jeortizquan, line 13: suggestion just as an idea use record instead? why a class?

src/main/java/eu/europeana/metis/sandbox/service/engine/BatchJobExecutor.java
jeortizquan, line 1: Suggestion this class is big, maybe is possible to split into smaller units.

jeortizquan (approved):
Is fine with me there are some observations and suggestions apply the ones that you consider they apply.

src/main/java/eu/europeana/metis/sandbox/batch/reader/AbstractIdentifiersItemReader.java
stzanakis: Dequeue is more performant that the list equivalient. 
In the current implementation there is no possibility for multithreading of this component and on a multi environment(which we also don't support yet), this in fact does not conflict since it's a per job singleton.

src/main/java/eu/europeana/metis/sandbox/batch/reader/AbstractIdentifiersItemReader.java
stzanakis: Well actually I think this is leftovers from before since the reading step is not run in parallel.
I will double check and remove it if so.

src/main/java/eu/europeana/metis/sandbox/batch/reader/AbstractIdentifiersItemReader.java
jeortizquan: my thought behind it: is it possible to scale up to multiple item readers?

src/main/java/eu/europeana/metis/sandbox/batch/repository/ExecutionRecordRepository.java
stzanakis: Excellent observation. Indeed this could be slower, I don't know how much and I don't know if it is too impactful. I cannot also really test it with the data in sandbox.
Nontheless I changed it to an equivalent JOIN query that should be in theory faster(I hope), but in any case I think we should just keep this in mind in case this is used with bigger dataset(it is not the case now).

src/main/java/eu/europeana/metis/sandbox/service/workflow/harvest/HarvestService.java
stzanakis: Sounds good, moved it to the common package.

src/main/java/eu/europeana/metis/sandbox/service/workflow/harvest/OaiHarvestService.java
stzanakis: Updated.

src/main/java/eu/europeana/metis/sandbox/service/workflow/harvest/FileHarvestTarget.java
stzanakis: Aaah yes! I had it as record but sonar started complaining about using the byte array in a custom equals and hashCode and toString. So I opted out to Lombok which also doesn't use it but avoids the sonar warning.
In any case, I reverted to record and added an explicit SuppressWarning. Not sure if there is an other way to do this tbh.

src/main/java/eu/europeana/metis/sandbox/service/engine/BatchJobExecutor.java
stzanakis: Hmm, it's on the limit I'd say. Not changing it now, but I'll see if I can think of something for the next ticket.

src/main/java/eu/europeana/metis/sandbox/batch/reader/AbstractIdentifiersItemReader.java
stzanakis: Generally no. Even the RepositoryItemReader from Spring batch is not run in parallel. The idea is that it gets a batch/chunk from the db and then sends the records to be processed in parallel.
Parallelization of this part could be problematic with paging and consistency.

PR #189, opened 2025-10-08
feat/MET-6819 Handle exception when step size bigger than dataset size

stzanakis:


src/main/java/eu/europeana/metis/sandbox/service/workflow/harvest/FileHarvestService.java
```diff
@@ -76,40 +77,41 @@ private HarvestIdentifiersResult<String> harvestIdentifiersFromCompressedArchive
...
-      recordLimitExceeded = recordCounter >= maxAllowedRecords;
-      if (recordLimitExceeded) {
-        log.warn("Reached maximum number of records ({}) during harvesting.", maxAllowedRecords);
+      if (isStepSizeBiggerThanDatasetSize(result.size(), currentIndex)) {
+        throw new StepIsTooBigException(currentIndex);
       }
     } catch (IOException e) {
       throw new ServiceException("Error harvesting records ", e);
     }
+    boolean recordLimitExceeded = result.size() >= maxAllowedRecords;
```
jeortizquan, line 107: should be this `result.size() > maxAllowedRecords;` ?
thinking on the upper boundary case.
before 
```
1000 >=1000    <=this should not be flagged as recordLimitExceeded
1001 >=1000    <=this is fine
```
after
```
1000 > 1000 => false
1001 > 1000 => true
```
what do you think?

stzanakis: The index starts from zero.

jeortizquan (approved):

src/main/java/eu/europeana/metis/sandbox/service/workflow/harvest/FileHarvestService.java
jeortizquan: :sweat_smile: true. got it.

PR #192, opened 2025-10-27
Feat/met 6840 file harvest with new extraction

stzanakis:
Requires https://github.com/europeana/metis-framework/pull/753 to be merged.

### DIFF
--- a/src/main/java/eu/europeana/metis/sandbox/batch/processor/listener/FileHarvestCleanupJobExecutionListener.java
+++ b/src/main/java/eu/europeana/metis/sandbox/batch/processor/listener/FileHarvestCleanupJobExecutionListener.java
@@ -1,0 +1,40 @@
+package eu.europeana.metis.sandbox.batch.processor.listener;
+
+import eu.europeana.metis.sandbox.service.workflow.harvest.FileHarvestService;
+import java.io.IOException;
+import java.nio.file.Path;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.io.FileUtils;
+import org.springframework.batch.core.JobExecution;
+import org.springframework.batch.core.JobExecutionListener;
+import org.springframework.batch.core.configuration.annotation.JobScope;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+/**
+ * Listener that deletes a specified temporary directory after the batch job finishes.
+ */
+@Slf4j
+@JobScope
+@Component
+public class FileHarvestCleanupJobExecutionListener implements JobExecutionListener {
+
+  @Value("#{jobParameters['datasetId']}")
+  private String datasetId;
+
+  @Override
+  public void beforeJob(JobExecution jobExecution) {
+    log.debug("File harvest job starting...");
+  }
+
+  @Override
+  public void afterJob(JobExecution jobExecution) {
+    Path pathToTempDestinationDirectoryById = FileHarvestService.getPathToTempDestinationDirectoryById(datasetId);
+    try {
+      FileUtils.deleteDirectory(pathToTempDestinationDirectoryById.toFile());
+    } catch (IOException e) {
+      log.error("Error deleting temporary directory {}", pathToTempDestinationDirectoryById, e);
+    }
+  }
+}
+
--- a/src/main/java/eu/europeana/metis/sandbox/batch/processor/listener/FileHarvestCleanupJobExecutionListener.java
+++ b/src/main/java/eu/europeana/metis/sandbox/batch/processor/listener/FileHarvestCleanupJobExecutionListener.java
@@ -29,7 +29,7 @@ public class FileHarvestCleanupJobExecutionListener implements JobExecutionListe
 
   @Override
   public void afterJob(JobExecution jobExecution) {
-    Path pathToTempDestinationDirectoryById = FileHarvestService.getPathToTempDestinationDirectoryById(datasetId);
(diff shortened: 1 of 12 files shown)


src/main/java/eu/europeana/metis/sandbox/batch/reader/FileIdentifiersItemReader.java
```diff
@@ -20,7 +21,7 @@
  */
 @Component
 @StepScope
-public class FileIdentifiersItemReader extends AbstractIdentifiersItemReader<String> {
+public class FileIdentifiersItemReader extends AbstractIdentifiersItemReader<Path> {
```
jeortizquan, line 25: :+1:

src/main/java/eu/europeana/metis/sandbox/batch/processor/listener/FileHarvestCleanupJobExecutionListener.java
```diff
@@ -0,0 +1,40 @@
...
+  @Override
+  public void beforeJob(JobExecution jobExecution) {
+    log.debug("File harvest job starting...");
+  }
+
+  @Override
+  public void afterJob(JobExecution jobExecution) {
+    Path pathToTempDestinationDirectoryById = FileHarvestService.getDeterministicPathToTempDirectoryById(datasetId);
+    try {
+      FileUtils.deleteDirectory(pathToTempDestinationDirectoryById.toFile());
```
jeortizquan, line 34: We need to see if this doesn't hung up in the server with big directories when deleting. if the container crashes I think the space will be freed, just consider cases where this cannot be cleaned up how to mitigate the risk.

src/main/java/eu/europeana/metis/sandbox/service/workflow/harvest/FileHarvestService.java
jeortizquan, line 1: the whole use of iterables in the project made me think to check if they are properly closed. Just double check them.

jeortizquan (approved):

src/main/java/eu/europeana/metis/sandbox/batch/processor/listener/FileHarvestCleanupJobExecutionListener.java
stzanakis: This is true, for bigger directories this will take longer to remove. This happens in Metis/ECloud already. But in metis-sandbox we are limiting the data size anyway.

src/main/java/eu/europeana/metis/sandbox/service/workflow/harvest/FileHarvestService.java
stzanakis: That is a good point, I have adapted the code slightly to clean the stream that is produced in some cases.

src/main/java/eu/europeana/metis/sandbox/batch/reader/AbstractIdentifiersItemReader.java
```diff
@@ -159,6 +164,20 @@ public ExitStatus afterStep(StepExecution stepExecution) {
...
     }
 
+    // Finally, close resources
+    if (harvestingIterator instanceof OaiHarvestingIterator) {
+      try {
+        harvestingIterator.close();
+      } catch (IOException e) {
+        log.error("Error closing OAI-PMH harvesting iterator", e);
+      }
+    }
```
jochen-vermeulen, line 175: Should this be more generic? Perhaps check if the harvesting iterator is instance of Closable? This is better maintainable (in case of new iterators or a change in existing iterators).

src/main/java/eu/europeana/metis/sandbox/batch/processor/listener/FileHarvestCleanupJobExecutionListener.java
jochen-vermeulen: Limiting in the sense of capping the size of the zip file?

src/main/java/eu/europeana/metis/sandbox/batch/reader/AbstractIdentifiersItemReader.java
stzanakis: The problem is that the HarvestingIterator close deletes the directory(for the file case), otherwise I wouldn't check it here at all and simply close any type.
So like in my todo comment in metis-framework in AbstractFileHarvestIterator, I think that the HarvestingIterator should not delete the provided directory path in the first place since it is not its responsibility.
Is that what you mean?

src/main/java/eu/europeana/metis/sandbox/batch/processor/listener/FileHarvestCleanupJobExecutionListener.java
stzanakis: Yes the amount of records in the file.

src/main/java/eu/europeana/metis/sandbox/batch/reader/AbstractIdentifiersItemReader.java
jochen-vermeulen: Something like that... So perhaps you should put a TODO here too, then? And create a follow-up ticket (if one does not already exists)?

stzanakis: Ok I have created the ticket and added the todo comment to investigate this as a separate concern.
https://europeana.atlassian.net/browse/MET-6889

PR #193, opened 2025-11-27
MET-6905 add update of enricher and dereferencer

jeortizquan:


src/main/java/eu/europeana/metis/sandbox/config/SandboxConfig.java
```diff
@@ -101,16 +103,26 @@ NormalizerFactory normalizerFactory() {
     return new NormalizerFactory();
   }
 
-  @Bean
-  EnrichmentWorker enrichmentWorker() throws DereferenceException, EnrichmentException {
+  @Bean(destroyMethod = "close")
```
stzanakis: I think the destroyMethod does not have to be declared if it is the "regular"/"default" known "close" method.

The doc says
This 'destroy method inference' is currently limited to detecting only public, no-arg methods named 'close' or 'shutdown'.

But of course it needs to be tested.

jeortizquan: :thinking: nice, we can simplify the code then.

stzanakis (approved):

PR #197, opened 2025-12-23
MET-6607 improve filesize detection and support redirects

jeortizquan:


src/main/java/eu/europeana/metis/sandbox/service/util/DatasetValidationService.java
```diff
@@ -58,4 +68,53 @@
...
+                             RequestConfig.custom()
+                                          .setRedirectsEnabled(true)
+                                          .setConnectTimeout(DEFAULT_CONNECT_TIMEOUT)
+                                          .setSocketTimeout(DEFAULT_SOCKET_TIMEOUT)
+                                          .build())
+                         .build()) {
+
+      HttpHead head = new HttpHead(urlString);
+
+      HttpResponse response = client.execute(head);
```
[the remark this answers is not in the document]
jeortizquan: follow up. https://europeana.atlassian.net/browse/MET-6975

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetExecutionService.java
```diff
@@ -130,12 +130,7 @@ public String createDatasetAndSubmitExecutionFile(DatasetMetadataRequest dataset
...
-    datasetValidationService.checkUrlContentLength(url);
-
-    try (InputStream inputStream = BoundedInputStream.builder()
-                                                     .setInputStream(new URI(url).toURL().openStream())
-                                                     .setMaxCount(datasetValidationService.getDefaultMaxFileSize().toBytes())
-                                                     .get()) {
+    try (InputStream inputStream = fileSizeValidationClient.download(URI.create(url))) {
       String filename = new URI(url).getPath();
       filename = filename.substring(filename.lastIndexOf('/') + 1);
       HttpHarvestParametersDTO harvestParametersDTO = new HttpHarvestParametersDTO(url, filename,
```
jochenvermeulen, line 146: The command `inputStream.readAllBytes()` (used to initialize this instance of `HttpHarvestParametersDTO`) now copies all bytes into a new byte array. You effectively now have the same data (that can be quite large) copied in memory twice. Can you think of a way to have it in memory only once?

src/main/java/eu/europeana/metis/sandbox/service/util/FileSizeValidationClient.java
```diff
@@ -0,0 +1,103 @@
...
+import org.springframework.http.ContentDisposition;
+import org.springframework.stereotype.Service;
+import org.springframework.util.unit.DataSize;
+import org.springframework.web.multipart.MaxUploadSizeExceededException;
+
+/**
+ * The type File size validation client.
+ */
+@Service
+public class FileSizeValidationClient extends AbstractHttpClient<URI, InputStream> {
```
jochenvermeulen: Is this the best name for this client? It does not just validate the file size, but also download the file (if it satisfies the file size requirement). Would a better name not be something like `ContentWithMaxSizeClient` or something like that?

src/main/java/eu/europeana/metis/sandbox/service/util/ContentUploadWithMaxSizeClient.java
```diff
@@ -0,0 +1,103 @@
...
+   * @param contentRetriever content retriever
+   * @return InputStream
+   * @throws IOException or MaxUploadSizeExceededException if the upload size is exceeded
+   */
+  @Override
+  protected InputStream createResult(URI providedURI, URI actualURI,
+      ContentDisposition contentDisposition, String mimetype, Long fileSize,
+      ContentRetriever contentRetriever) throws IOException {
+    if (fileSize != null && fileSize > defaultMaxFileSize.toBytes()) {
+      throw new MaxUploadSizeExceededException(defaultMaxFileSize.toBytes());
```
jochenvermeulen: I don't think using `MaxUploadSizeExceededException` is right here... this exception applies if a client is uploading data. A client (and the spring framework) may misunderstand this as they are not actually uploading content. 

However, if the UI already expects this exception (and this error code 413) then let's not change it.

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetExecutionService.java
```diff
@@ -147,6 +142,10 @@ public String createDatasetAndSubmitExecutionHttp(DatasetMetadataRequest dataset
       batchJobExecutor.execute(executionMetadata);
       return executionMetadata.getDatasetMetadata().getDatasetId();
     } catch (IOException | URISyntaxException e) {
+      if (e instanceof IOException ioException && ioException.getMessage().contains("Status code 404")) {
+        throw new ServiceException(HARVESTING_ERROR_MESSAGE,
+            new FileNotFoundException("The provided URL does not exist or cannot be accessed: " + url));
+      }
```
jochenvermeulen: This I don't understand. First of all, I think it is bad practice to check the exception message like this: it may change in the future without any warning. Second: what is gained by creating this new exception for this specific case? Why not for all IO exceptions? 

My proposal: for ALL IO exceptions: there clearly is a problem retrieving the data. Instead, use the Spring exception conforming to the 400 (Bad request) or, perhaps better, the 502 (bad gateway).

src/main/java/eu/europeana/metis/sandbox/service/util/FileSizeValidationClient.java
jeortizquan: I see, is a matter of naming. I will change it to `ContentUploadWithMaxSizeClient`

src/main/java/eu/europeana/metis/sandbox/service/util/ContentUploadWithMaxSizeClient.java
jeortizquan: I am changing the name of the class and hope this name makes more sense and the use of this exception as well.

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetExecutionService.java
jeortizquan: done. simplified.

src/main/java/eu/europeana/metis/sandbox/service/dataset/DatasetExecutionService.java
jeortizquan: this comes from the AbstractHttpClient, and was a way to wrap the file not found in an url, so in the UI was an specific message. I added a method for verbosity, see if this is better.

PR #198, opened 2026-01-14
Debt/met 6557 spring upgrade

stzanakis:


src/main/java/eu/europeana/metis/sandbox/config/webmvc/WebMvcConfig.java
```diff
@@ -2,6 +2,7 @@
 
 import eu.europeana.metis.sandbox.controller.ratelimit.RateLimitInterceptor;
 import org.apache.commons.lang3.ArrayUtils;
+import org.jspecify.annotations.NonNull;
```
jeortizquan: is this intended to be like this?
was this one `jakarta.validation.constraints.NotNull` or `org.jetbrains.annotations.NotNull`

stzanakis: Removed, since it's a local warning(perhaps a bug)

jeortizquan (approved):

PR #202, opened 2026-05-07
bug/MET-7094_add_support_error_counting_debias

jeortizquan:
As consequence of using new metis-sandbox engine. Error count where not considered in the calculation of COMPLETE status.

### DIFF
--- a/.run/metis-sandbox (local).run.xml
+++ b/.run/metis-sandbox (local).run.xml
@@ -1,6 +1,12 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="metis-sandbox (local)" type="SpringBootApplicationConfigurationType" factoryName="Spring Boot" folderName="spring-run-configuration">
     <option name="ALTERNATIVE_JRE_PATH" value="temurin-21" />
+    <envs>
+      <env name="logging.config" value="file:///data/metis-configuration/k8s/common-components/log4j2-xml/log4j2.xml" />
+      <env name="spring.config.location" value="file:///data/metis-configuration/metis-sandbox/metis-sandbox-rest/k8s/overlays/local/components/properties/application.yml" />
+      <env name="truststore.path" value="/data/metis-configuration/k8s/common-components/custom-truststore-jks/custom-truststore.jks" />
+      <env name="server.port" value="8081" />
+    </envs>
     <module name="metis-sandbox-rest" />
     <option name="PROGRAM_PARAMETERS" value="--spring.config.location=file:///data/metis-configuration/metis-sandbox/metis-sandbox-rest/k8s/overlays/local/components/properties/application.yml --server.servlet.context-path=/metis-sandbox --logging.config=file:///data/metis-configuration/k8s/common-components/log4j2-xml/log4j2.xml --truststore.path=/data/metis-configuration/k8s/common-components/custom-truststore-jks/custom-truststore.jks" />
     <option name="SPRING_BOOT_MAIN_CLASS" value="eu.europeana.metis.sandbox.SandboxApplication" />
--- a/.run/metis-sandbox (local).run.xml
+++ b/.run/metis-sandbox (local).run.xml
@@ -1,12 +1,6 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="metis-sandbox (local)" type="SpringBootApplicationConfigurationType" factoryName="Spring Boot" folderName="spring-run-configuration">
     <option name="ALTERNATIVE_JRE_PATH" value="temurin-21" />
-    <envs>
-      <env name="logging.config" value="file:///data/metis-configuration/k8s/common-components/log4j2-xml/log4j2.xml" />
-      <env name="spring.config.location" value="file:///data/metis-configuration/metis-sandbox/metis-sandbox-rest/k8s/overlays/local/components/properties/application.yml" />
-      <env name="truststore.path" value="/data/metis-configuration/k8s/common-components/custom-truststore-jks/custom-truststore.jks" />
-      <env name="server.port" value="8081" />
-    </envs>
     <module name="metis-sandbox-rest" />
     <option name="PROGRAM_PARAMETERS" value="--spring.config.location=file:///data/metis-configuration/metis-sandbox/metis-sandbox-rest/k8s/overlays/local/components/properties/application.yml --server.servlet.context-path=/metis-sandbox --logging.config=file:///data/metis-configuration/k8s/common-components/log4j2-xml/log4j2.xml --truststore.path=/data/metis-configuration/k8s/common-components/custom-truststore-jks/custom-truststore.jks" />
     <option name="SPRING_BOOT_MAIN_CLASS" value="eu.europeana.metis.sandbox.SandboxApplication" />
--- a/metis-sandbox-service/src/main/java/eu/europeana/metis/sandbox/service/debias/DeBiasStateService.java
+++ b/metis-sandbox-service/src/main/java/eu/europeana/metis/sandbox/service/debias/DeBiasStateService.java
@@ -112,13 +112,12 @@ public class DeBiasStateService {
       debiasState = DebiasState.INVALID;
     }
 
-    if (datasetDeBiasEntity == null) {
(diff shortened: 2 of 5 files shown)


.run/metis-sandbox (local).run.xml
```diff
@@ -1,6 +1,12 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="metis-sandbox (local)" type="SpringBootApplicationConfigurationType" factoryName="Spring Boot" folderName="spring-run-configuration">
     <option name="ALTERNATIVE_JRE_PATH" value="temurin-21" />
+    <envs>
```
stzanakis: How did this come here? They are defined below.

metis-sandbox-service/src/main/java/eu/europeana/metis/sandbox/service/debias/DeBiasStateService.java
```diff
@@ -88,9 +95,14 @@ public DeBiasStatusDTO getDeBiasStatus(String datasetId) {
...
-
+    long totalRecordsDebiasError = executionRecordErrorRepository.countByExecutionRun_DatasetIdAndExecutionRun_ExecutionName(
+        datasetId,
+        BatchJobType.DEBIAS.name());
+    log.info(
+        "Debias status for datasetId: [{}], totalRecordsToDebias: [{}], totalRecordsDebiased: [{}], totalRecordsDebiasError: [{}]",
+        datasetId, totalRecordsToDebias, totalRecordsDebiased, totalRecordsDebiasError);
     final DebiasState debiasState;
-    if (totalRecordsToDebias > 0 && (totalRecordsToDebias == totalRecordsDebiased)) {
+    if (totalRecordsToDebias > 0 && (totalRecordsToDebias == totalRecordsDebiased + totalRecordsDebiasError)) {
```
stzanakis: With that change the 3rd if needs to be updated as well so that it is in PROCESSING state when either of the counters is above zero.
Since the addition of the two values(totalRecordsDebiased + totalRecordsDebiasError) is used in two places, we can create a variable to capture them above.

metis-sandbox-service/src/test/java/eu/europeana/metis/sandbox/service/debias/DeBiasStateServiceTest.java
```diff
@@ -222,6 +226,59 @@ void getDeBiasReport() {
...
+        List.of(recordDeBiasMainEntity));
+    when(recordDeBiasDetailRepository.findByDebiasIdId(recordDeBiasMainEntity.getId())).thenReturn(
+        List.of(recordDeBiasDetailEntity));
+
+    DeBiasReportDTO deBiasReportDTO = deBiasStateService.getDeBiasReport(datasetId);
+
+    assertEquals(DebiasState.COMPLETED, deBiasReportDTO.getDebiasState());
+    assertEquals(nowDate, deBiasReportDTO.getCreationDate());
+    assertEquals(10, deBiasReportDTO.getTotal());
+    assertEquals(8, deBiasReportDTO.getProcessed());
```
stzanakis: This is not correct anymore.
The DeBiasStatusDTO in the DeBiasStateService should be updated with the value (totalRecordsDebiased + totalRecordsDebiasError), as suggested there a new field.
It should be 10 here.

pom.xml
```diff
@@ -33,8 +33,8 @@
     <!--Internal-->
     <version.metis>17-SNAPSHOT</version.metis>
     <version.metis-common>19-SNAPSHOT</version.metis-common>
```
stzanakis: You can update this to 20-SNAPSHOT

pom.xml
```diff
@@ -33,8 +33,8 @@
     <!--Internal-->
     <version.metis>17-SNAPSHOT</version.metis>
```
stzanakis: You can also update this to 18-SNAPSHOT

stzanakis (changes requested):

.run/metis-sandbox (local).run.xml
jeortizquan: :thinking: yes. I see. cleaned up. done.

metis-sandbox-service/src/test/java/eu/europeana/metis/sandbox/service/debias/DeBiasStateServiceTest.java
jeortizquan: ah yes. I see, in the follow up ticket we can introduce the proper field. I updated the code as described. done.

pom.xml
jeortizquan: done.

pom.xml
jeortizquan: done.

metis-sandbox-service/src/main/java/eu/europeana/metis/sandbox/service/debias/DeBiasStateService.java
jeortizquan: good one. thanks. refactor done.

metis-sandbox-service/src/main/java/eu/europeana/metis/sandbox/service/debias/DeBiasStateService.java
```diff
@@ -88,25 +95,30 @@ public DeBiasStatusDTO getDeBiasStatus(String datasetId) {
...
+        BatchJobType.DEBIAS.name());
+    log.info(
+        "Debias status for datasetId: [{}], totalRecordsToDebias: [{}], totalRecordsDebiased: [{}], totalRecordsDebiasError: [{}]",
+        datasetId, totalRecordsToDebias, totalRecordsDebiased, totalRecordsDebiasError);
     final DebiasState debiasState;
-    if (totalRecordsToDebias > 0 && (totalRecordsToDebias == totalRecordsDebiased)) {
+    final long totalDebiasProcessed = totalRecordsDebiased + totalRecordsDebiasError;
+    if (totalRecordsToDebias > 0 && (totalRecordsToDebias == totalDebiasProcessed)) {
       debiasState = DebiasState.COMPLETED;
     } else if (totalRecordsToDebias >= 0 && totalRecordsDebiased == 0) {
```
stzanakis: I think this one needs to be updated as well. It should be in READY state when nothing is processed yet(even errrors)

metis-sandbox-service/src/main/java/eu/europeana/metis/sandbox/service/debias/DeBiasStateService.java
```diff
@@ -88,25 +95,30 @@ public DeBiasStatusDTO getDeBiasStatus(String datasetId) {
...
     }
 
-    if (datasetDeBiasEntity == null) {
-      return new DeBiasStatusDTO(Integer.valueOf(datasetId), debiasState, ZonedDateTime.now(), totalRecordsToDebias,
-          totalRecordsDebiased);
-    } else {
-      return new DeBiasStatusDTO(Integer.valueOf(datasetId), debiasState,
-          datasetDeBiasEntity.getCreatedDate(), totalRecordsToDebias, totalRecordsDebiased);
+    ZonedDateTime creationDate = ZonedDateTime.now();
+    if (datasetDeBiasEntity != null && datasetDeBiasEntity.getCreatedDate() != null) {
```
stzanakis: This changes semantics.
If the entity exists but the creation date is null it is then overwritting the null with the `now` which now is not the creation date anymore and also will potentially hide a problem in another place in the code.

stzanakis (changes requested):

metis-sandbox-service/src/main/java/eu/europeana/metis/sandbox/service/debias/DeBiasStateService.java
jeortizquan: true. I missed that. I will allow the null.

metis-sandbox-service/src/main/java/eu/europeana/metis/sandbox/service/debias/DeBiasStateService.java
jeortizquan: o gotcha.

stzanakis (approved):

PR #205, opened 2026-06-30
feat/MET-7160_rename_publish_to_preview_for_index_context

stzanakis:


.github/workflows/perf.yml
```diff
@@ -2,9 +2,6 @@ name: performance
 
 on:
   workflow_dispatch:
```
jeortizquan, line 4: was sonar complaining of this?
```
 branches:
      - feat/*
      - develop
```
this allows to use test branches

metis-sandbox-common/src/main/java/eu/europeana/metis/sandbox/common/batch/FullBatchJobType.java
```diff
@@ -70,6 +70,6 @@ public static FullBatchJobType validateAndGetFullBatchJobType(String prefix, Str
    * organized.
    */
   public enum BatchJobGroup {
-    HARVEST, CURATE, INDEX, DEPUBLISH, DEBIAS
+    HARVEST, CURATE, INDEX, DEBIAS
```
jeortizquan, line 73: why remove `DEPUBLISH`? or add it back when is needed?

metis-sandbox-service/src/main/java/eu/europeana/metis/sandbox/dto/debias/DeBiasStatusDTO.java
```diff
@@ -1,15 +1,16 @@
...
-import java.time.ZonedDateTime;
+import java.time.Instant;
+import lombok.Getter;
 
 /**
  * The type Detection info dto.
  */
 @Schema(name = DeBiasStatusDTO.SWAGGER_MODEL_NAME)
+@Getter
 public class DeBiasStatusDTO {
```
jeortizquan, line 16: suggestion to add `@AllArgsConstructor` and remove the current one.

.github/workflows/perf.yml
stzanakis: Yes, the schema has changed so I add the restriction on the run with an if.

metis-sandbox-common/src/main/java/eu/europeana/metis/sandbox/common/batch/FullBatchJobType.java
stzanakis: Yes, we'll re-add it, since we are not yet sure if we are going to implemented.

metis-sandbox-service/src/main/java/eu/europeana/metis/sandbox/dto/debias/DeBiasStatusDTO.java
stzanakis: done.

jeortizquan (approved):

